Issue #9, opened 2021-09-07
Configure Frontend Linter


willyyhuang:
Linting will be done for frontend files using [Prettier](https://prettier.io/) and [ESLint](https://eslint.org/). This will make sure JavaScript files pass static analysis checks and follow coding guidelines.

PR #10, opened 2021-09-07
chore: #9 configure front end linter

willyyhuang:
Add
- Eslint
- Prettier
- Babel
- Webpack config
- Typescript config


docker-compose.yml
```diff
@@ -5,47 +5,47 @@ services:
     volumes:
       - .:/var/www/html/
     ports:
-      - "80:80"
```
willyyhuang, line 8: @m-triassi can you tell me if this will affect the file, looks like prettier went a little too ham

PR #16, opened 2021-09-08
Add CI pipeline via GitHub Actions

### LINKED ISSUE #3: Configure CodeClimate

### LINKED ISSUE #8: Create GitHub Actions

m-triassi:
## Description
Closes #8 
Closes #3 

Add a CI Pipeline which sets up the entire project, and runs auxiliary, required services (such as MySQL), builds the front-end, and then executes the PHP Unit Test suite. 

Additionally adds a step that hooks into CodeClimates test reporter and passes generate Clover reports.


.github/workflows/test-suite.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Cache dependencies
+        uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache/files
+          key: dependencies-laravel-6.x-php-7.4-composer-${{ hashFiles('composer.json') }}
```
pascwhale, line 33: We should consider the following changes to the label.
```suggestion
          key: dependencies-laravel-8.x-php-8.0-composer-${{ hashFiles('composer.json') }}
```

.github/workflows/test-suite.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Configure Composer
+        run: |
+          composer install -n --no-ansi --no-interaction --no-scripts --no-progress --prefer-dist
+
+      - name: Setup Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          cache: yarn
```
pascwhale, line 50: We should also consider caching this like we do for composer.

pascwhale (changes requested):
Please check my comments.

.github/workflows/test-suite.yml
m-triassi: Ah whoops missed that!

.github/workflows/test-suite.yml
m-triassi: Good call.

m-triassi: Actually after taking a second look, it seems like this action follows what actions/cache@v1 does and caches the global dependencies, See [here](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies) for more info

m-triassi:
Just realized i asked for a review a little prematurely, we still need to set up cypress as well! Hold off on reviewing until then.

m-triassi:
Looks like cypress is running properly in the Workflow now 🎉

pascwhale (approved):
Great, I'm glad this is set up!

PR #17, opened 2021-09-08
Add and setup cypress to the project

### LINKED ISSUE #7: Set Up Cypress

luigibesani:
## Description
Closes #7
Install cypress and configure
Add cypress through composer
Add cypress through yarn


m-triassi (approved):
This look alright, we'll probably want to consider removing the example integration tests, but they can remain for now to give us something to test with. Nice work!

PR #20, opened 2021-09-09
Add and setup psalm and add it to CI pipeline

### LINKED ISSUE #6: Set Up Psalm
[Psalm is a free & open-source static analysis tool that helps you identify problems in your code, so you can sleep a little better.](https://psalm.dev/)

RobertoNittolo:
## Description
Closes #6 

Add Psalm and setup to the repo.
Add Psalm to the CI pipleine with Github actions

*** To run Psalm locally run the following command `./vendor/bin/psalm`

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -82,11 +82,6 @@ jobs:
 # Second job that runs Psalm
 Psalm
 
-on:
-  pull_request:
-    branches: [ main ]
-  push:
-    branches: [ main ]
   Psalm:
     name: psalm
     runs-on: ubuntu-latest
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -78,3 +78,41 @@ jobs:
         uses: cypress-io/github-action@v2
         with:
           config: baseUrl=http://127.0.0.1:8000
+
+# Second job that runs Psalm
+name: Psalm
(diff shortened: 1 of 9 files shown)


.github/workflows/test-suite.yml
```diff
@@ -78,3 +78,31 @@ jobs:
...
+        with:
+          php-version: 8.0
+          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, bcmath, soap, intl, gd, exif, iconv, imagick
+          coverage: xdebug
+
+      - name: Cache dependencies
+        uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache/files
+          key: dependencies-laravel-8.x-php-8.0-composer-${{ hashFiles('composer.json') }}
```
RobertoNittolo, line 101: Do I need to also cache the dependencies if it is already being cached in the Code-Climate-tests pipeline?

m-triassi: I believe you do, seeing as the code climate and psalm pipelines run on different virtual machines -- no harm either way

m-triassi (approved):
From my perspective, this looks just about perfect. As an aside, please don't prefix branch names with a "#" it will cause problems when running commands in a terminal; simply using `6-set-up-psalm` would be sufficient

PR #50, opened 2021-09-20
Component library and folder structure setup

### LINKED ISSUE #48: Setup component library, typescript and webpack configs

willyyhuang:
Closes #48.


m-triassi (commented):
Is there a reason we deleted the entire `Components` directory? Where will the reusable components go?

willyyhuang:
> Is there a reason we deleted the entire `Components` directory? Where will the reusable components go?

it will go under resources/components, now there's no components therefore git did not commit the folder, will be added back once we have written re-usable components

webpack.mix.js
```diff
@@ -11,13 +11,13 @@ const mix = require('laravel-mix')
...
-    .postCss('resources/css/app.css', 'public/css', [
-        require('postcss-import'),
-        require('tailwindcss'),
-        require('autoprefixer'),
-    ])
+    // .postCss('resources/css/app.css', 'public/css', [
+    // require('postcss-import'),
+    // require('tailwindcss'),
+    // require('autoprefixer'),
+    // ])
```
m-triassi, line 20: We can probably delete these lines, outright, if we don't intend to use tailwind, or any other customized CSS packages
```suggestion
```

m-triassi (approved):
Great stuff, we'll want to keep an eye on the increased install / compile times for some of the front ends, but otherwise this perfect. 

I left 1 comment, but I'll leave it up to you if you want to address it :)

webpack.mix.js
willyyhuang: let's leave it for now, maybe we will define a css package later so this would be somewhat useful

PR #61, opened 2021-09-24
Scaffold basic backend structure

### LINKED ISSUE #42: Create Character Model
## Description
The character model contains metadata about a player's character:
* Season
* Race (such as elf, dwarf, orc, human, etc)
* Classes
* Background
* Lifestyle
* Current Level
* Total Gold Pieces
* Total Downtime
* Faction
* Renown Rank
* Total Renown
* Privacy setting

### LINKED ISSUE #44: Implement Show Character Endpoint

### LINKED ISSUE #46: Implement Update Character Endpoint

### LINKED ISSUE #47: Implement Destroy Character Endpoint

### LINKED ISSUE #60: Install Blueprint and create project scaffolding
## Description

Leveraging Blueprint, we should translate the created class diagram and ERD to a manifest that can then be used to generate all the appropriate files we will need. This ensures we will be able to focus on the truly custom functionality of the project rather than writing repetative boilerplate code.

## Possible Implementation

1. install Laravel-Blueprint
2. write a manifest describing all the models, relationships, and controllers the platform
3. run the blueprint builder
4. verify all of the created files are somewhat correct
5. add a "dev" route specifically for viewing intertia pages, as this wont be covered by blueprint

Some notes about this approach: the inital return of the controllers will likely be incorrect considering blueprint is not aware of our use of InteriaJS, we can progressively use the correct return values as the project progresses

## Out of scope

N/a

- The site should still load properly after running the builder
- There should not be any errors thrown when running CLI commands such as `php artisan migrate`
- The created routes should generally be valid

m-triassi:
## Description
Closes #60 
Closes #42
Closes #44 
Closes #46 
Closes #47 

Adds all migrations,  models, and controllers, according to architecture specification documents (ERD, Class Diagram, Domain Model).

Registers routes for newly created resource controllers.

Additionally, adds tests covering all controllers, defines preliminary model relationships, and resource request validation.

Installs Blueprint, which was used to generate the aforementioned files.

Omits basic blade views, since this platform will instead be using InteriaJS + React. 

## How to test
1. Ensure that all endpoints work, via Postman or Insomnia
   1. Get example collection for insomnia here: [SessionTome_Insomnia.zip](https://github.com/DnD-Montreal/session-tome/files/7229072/SessionTome_Insomnia.zip)
2. run test suite locally by running `php artisan test --env=env`


m-triassi:
@IanjhPhillips @pascwhale you two are probably good to give this a review now, all thats missing is satisfying CodeClimate's Diff coverage threshold -- @JasonGadoury will be sorting that out by tonight.

app/Http/Requests/AdventureStoreRequest.php
```diff
@@ -0,0 +1,32 @@
...
+class AdventureStoreRequest extends FormRequest
+{
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
+        return true;
```
IanjhPhillips, line 16: should these be default to true?

IanjhPhillips (commented):
This should have been implemented across multiple pull requests.

Otherwise, looks good to me,

app/Models/Character.php
```diff
@@ -0,0 +1,61 @@
...
+    public function items()
+    {
+        return $this->hasMany(\App\Models\Item::class);
+    }
+
+    public function trades()
+    {
+        return $this->hasMany(\App\Models\Trade::class);
+    }
+}
```
pascwhale, line 61: Is a function missing to represent a relationship with entries?

app/Http/Controllers/CharacterController.php
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * @param \App\Http\Requests\CharacterUpdateRequest $request
+     * @param \App\Models\Character $character
+     * @return \Illuminate\Http\Response
+     */
+    public function update(CharacterUpdateRequest $request, Character $character)
+    {
+        $character->update($request->validated());
+
+        $request->session()->flash('character.id', $character->id);
```
pascwhale, line 76: What is the purpose of the [flash](https://laravel.com/docs/8.x/session#flash-data) function here? Is it updating the object only for the next request?

database/factories/CharacterFactory.php
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @return array
+     */
+    public function definition()
+    {
+        return [
+            'user_id' => User::factory(),
+            'name' => $this->faker->name(),
+            'race' => $this->faker->word(),
+            'class' => $this->faker->word(),
```
pascwhale, line 30: Is there a way to create values for these fields from an `enum`?

database/factories/CharacterFactory.php
```diff
@@ -0,0 +1,37 @@
...
+            'level' => $this->faker->word(),
```
pascwhale, line 31: Was there a limit to the level a character can reach?

tests/Feature/Http/Controllers/CharacterControllerTest.php
```diff
@@ -0,0 +1,195 @@
...
+ * @see \App\Http\Controllers\CharacterController
+ */
+class CharacterControllerTest extends TestCase
+{
+    use AdditionalAssertions;
+    use RefreshDatabase;
+    use WithFaker;
+
+    /**
+     * @test
```
pascwhale, line 22: I think these should eventually have better descriptions.

app/Models/Character.php
m-triassi: Actually, you're right; the relationship is missing, ill add that!

app/Http/Controllers/CharacterController.php
m-triassi: The purpose of this is so that the front end can easily track precisely which object has change, by accessing it via the session. This helps with state management

database/factories/CharacterFactory.php
m-triassi: Yes, there is, but iirc we resolved that since there is the possibility that users can have special races, we left this as strings. 

Classes were left as strings because characters can do something called "multi-classing" which means they may have an an arbitrary amount of classes on a character

database/factories/CharacterFactory.php
m-triassi: There is! They can only be up to level 20, we can fix that.

tests/Feature/Http/Controllers/CharacterControllerTest.php
m-triassi: I think we'll do a deeper documentation pass at some point, just the Doc annotation is sufficient for now

app/Http/Requests/AdventureStoreRequest.php
m-triassi: This one in particular, yes, since this is just a place holder for now.

app/Models/Character.php
pascwhale: I see that you added the relationship, thanks!

database/factories/CharacterFactory.php
pascwhale: I didn't notice that the level was created as a word. Thanks for changing that too.

pascwhale (approved):
Everything looks good to me.

IanjhPhillips (approved):

m-triassi:
Looks like we're good to merge this! 🚀

PR #77, opened 2021-09-29
49 application layout

### LINKED ISSUE #49: Application layout

luigibesani:
Closes #49


.gitignore
```diff
@@ -3,6 +3,7 @@
 /public/storage
 /public/js
 /public/css
+/public/images
```
willyyhuang, line 6: not sure if this is to be ignored, does webpack compile images here? @m-triassi

resources/components/Layout/AppLayout.tsx
```diff
@@ -1,7 +1,70 @@
 import React from 'react'
+import Grid from '@mui/material/Grid'
+import SVG from 'react-inlinesvg'
+import {Avatar, Link} from '@mui/material'
```
willyyhuang, line 4: you can import `Grid` together here i think

resources/components/Layout/AppLayout.tsx
```diff
@@ -1,7 +1,70 @@
...
+import associationLogo from '../../icons/ParentAssociationLogoWhite.svg'
```
willyyhuang, line 5: maybe rename this to like `DNDLogo` or something, parent association is pretty vague

.gitignore
m-triassi: @willyyhuang It does, yeah, images we want to keep in the repo (say for icons and such) should be live in `resources`

m-triassi:
I know this is nit picky, but @luigibesani before this PR leaves draft please choose a more descriptive title, rather than an auto generated one. Something like "Add basic application layout". Additionally its always good to describe what this PR does in the body so a reviewer will know more or less what to look for in the app

resources/components/Layout/AppLayout.tsx
luigibesani: You're right about this! I renamed it to an appropriate name.

resources/components/Layout/AppLayout.tsx
luigibesani: Thank you for pointing that out. 👍

PR #82, opened 2021-10-09
Add Jest to code climate pipeline

### LINKED ISSUE #62: Add Jest to CI/CD pipeline
## Description

Jest coverage reports should be sent over to code climate so that it can assess what files are covered by tests, to what degree, as well as how maintainable the files are.

## Possible Implementation

This entails running the Jest suite within our pipeline and generating a clover coverage report that is then passed to CodeClimate

## Out of scope

N/a

Coverage report must provide data to Code Climate

RobertoNittolo:
##Description
Closes #62
Adds Jest to the pipeline
Runs Jest tests 
Generates clover coverage report and sends the report to code climate

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -55,15 +55,16 @@ jobs:
           yarn prod
 
       - name: Run PHPUnit
+        env:
+          DB_PORT: ${{ job.services.mysql.ports[3306] }}
         run: ./vendor/bin/phpunit --coverage-clover tests/coverage.xml
 
       - name: Run Jest
         run: yarn jest --coverage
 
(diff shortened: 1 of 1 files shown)


m-triassi:
> 
> 
> @m-triassi is there any way that I can verify that both PHPUnit and Jests code coverage reports were merged and published to Code Climate from my branch before my changes get merged into main?

So it looks like there isn't a way to actually look at the new coverage for files before merging them; but im pretty confident we're receiving both coverage reports since the CodeClimate bot is reporting the total coverage for the project will rise after this PR. 
> This pull request will bring the total coverage in the repository to 90.1% (0.5% change).

Since you didnt add tests in this PR i can only assume that change is from the coverage granted by Jest :)

PR #89, opened 2021-10-14
Adjust API Access control

### LINKED ISSUE #83: Reconfigure login to work for SPAs
## Description

Currently the login and register routes are returning 302s which is not ideal. Instead it should return an (intertia) response.
Additonally it was found in #80 there are issues with persisting sessions from logins, this should be fixed.

## Possible Implementation

Swap return statements on login and register routes for intertia responses. Look into partial reloads vs plain ajax responses

## Out of scope

N/A

The login should be able to be interaced with asynchronously

m-triassi:
## Description
Closes #83 

Protects all resource controller routes under Auth Middleware. Enables sanctum middleware for stateful Front end requests.
Swaps out redirect in login / register for Authenticated user model.

_note:_ There isn't really much to test here, so i'm omitting the "How to Test" portion


m-triassi:
@willyyhuang Heads up for you, in the end i changed the `login` endpoint to return the user, same with `register` since inertia partial reloads was a little janky

m-triassi:
Ouf Test suite is failing hard 😭

m-triassi:
Ill be fixing the rest of the failing tests tomorrow, likely

m-triassi:
@IanjhPhillips Each test wont strictly need to implement a setup method, though i imagine that it would make passing tests easier if the test pertains to a controller behind the auth middleware. Its mostly a case by case basis

PR #144, opened 2021-10-30
DM Entry Controller

### LINKED ISSUE #117: Implement DM Entry Creation
## Description
Implement logic for creating DM entries

## Possible Implementation
Create DM entry controller
Add controller to route list
Implement store function on controller

IanjhPhillips:
## Description
Closes #117
- Creates a controller to handle specific behaviour of DM entries which do not belong to a character.
- Upon /entry.store , if type is 'dm' the response will be a redirect to the index of DMEntryController.
- A new view may be needed to be added in the future in case we wish to differentiate listing characterless DM entries from a list of entries on a character.
- Also fixes an issue in web.php routes where some routes did not have middleware.

### DIFF
--- a/app/Observers/EntryObserver.php
+++ b/app/Observers/EntryObserver.php
@@ -14,10 +14,12 @@ class EntryObserver
      */
     public function created(Entry $entry)
     {
-        $character = $entry->character;
-        $character->level += $entry->levels;
+        if ($entry->type != Entry::TYPE_DM) {
+            $character = $entry->character;
+            $character->level += $entry->levels;
 
-        $character->save();
+            $character->save();
+        }
     }
 
     /**
@@ -28,11 +30,13 @@ class EntryObserver
      */
     public function updated(Entry $entry)
     {
-        $character = $entry->character;
-        if ($entry->isDirty('levels')) {
-            $levelDelta = $entry->levels - $entry->getOriginal('levels');
-            $character->level += $levelDelta;
-            $character->save();
+        if ($entry->type != Entry::TYPE_DM) {
+            $character = $entry->character;
+            if ($entry->isDirty('levels')) {
+                $levelDelta = $entry->levels - $entry->getOriginal('levels');
+                $character->level += $levelDelta;
+                $character->save();
+            }
         }
     }
 
@@ -44,10 +48,12 @@ class EntryObserver
      */
     public function deleting(Entry $entry)
     {
-        $character = $entry->character;
-        if ($entry->levels) {
-            $character->level -= $entry->levels;
-            $character->save();
+        if ($entry->type != Entry::TYPE_DM) {
+            $character = $entry->character;
+            if ($entry->levels) {
+                $character->level -= $entry->levels;
+                $character->save();
+            }
         }
     }
 }
--- a/app/Observers/EntryObserver.php
+++ b/app/Observers/EntryObserver.php
@@ -35,6 +35,11 @@ class EntryObserver
                 $levelDelta = $entry->levels - $entry->getOriginal('levels');
                 $character->level += $levelDelta;
                 $character->save();
+            } elseif ($entry->isDirty('character_id')) {
+                $character->level = (is_null($character->level))
+                    ? $entry->levels
+                    : $character->level + $entry->levels;
+                $character->save();
             }
         }
     }
--- a/app/Observers/EntryObserver.php
+++ b/app/Observers/EntryObserver.php
@@ -14,10 +14,9 @@ class EntryObserver
      */
     public function created(Entry $entry)
     {
-        if ($entry->type != Entry::TYPE_DM) {
+        if ($entry->type != Entry::TYPE_DM || !is_null($entry->character)) {
             $character = $entry->character;
             $character->level += $entry->levels;
-
             $character->save();
         }
     }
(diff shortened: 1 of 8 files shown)


pascwhale (commented):
Hey @IanjhPhillips,

The `Possible Implementation` section in the issue (#117) lists three steps:
- Create DM entry controller
- Add controller to route list
- Implement store function on controller

Is it me or is the third step not there? I'm wondering if you didn't need that step for the issue?
The concept of [Partial Resource Routes](https://laravel.com/docs/8.x/controllers#restful-partial-resource-routes) is also new to me, but the added line in the `routes/web.php` file only shows `index` for the new route.

```
Route::resource('dm-entry', \App\Http\Controllers\DMEntryController::class)
        ->only('index');
```

pascwhale:
I'm also struggling to see where we can add tests to meet the diff coverage threshold set by CodeClimate. The ones you have seem to cover the index you added.

IanjhPhillips:
I realised that theres nothing fundamentally different about creating a DM entry vs a regular entry, so the actual creation can be handled by the EntryController store function, whereas the index might be slightly different as you're specifically only getting entries with type 'dm', so I chose to have a redirect case wthin entry.store which redirects to dm-entry.index if the store request was for a dm-entry.

This way, we only need one form for creating entries, which can be either attached to a player or a free floating DM entry.

IanjhPhillips:
I have some ideas where test coverage can be added.

app/Observers/EntryObserver.php
```diff
@@ -14,10 +14,12 @@ class EntryObserver
      */
     public function created(Entry $entry)
     {
-        $character = $entry->character;
-        $character->level += $entry->levels;
+        if ($entry->type != Entry::TYPE_DM) {
```
pascwhale, line 17: I think we should have a better check for the observer functions. Once we have entries attached to a character, we might want levels ups to be tracked if the player chooses to level up using that entry.

tests/Feature/Http/Controllers/EntryControllerTest.php
```diff
@@ -227,6 +227,21 @@ public function update_redirects()
...
+            'dungeon_master_id' => $dungeon_master_user->id,
+            'dungeon_master' => $dungeon_master,
+            'date_played' => $date_played,
+            'location' => $location,
+            'type' => $type,
+            'levels' => $levels,
+            'gp' => $gp,
+        ]);
+
+        $response->assertForbidden();
```
pascwhale, line 243: I don't think I understand why you're expecting this response to fail. Doesn't the entry update keep the last character attached to it?

IanjhPhillips: There's no character_id in the *request body* and type is not 'dm' so it fails authorization in EntryUpdateRequest.

app/Observers/EntryObserver.php
IanjhPhillips: If an entry specifies a level, the character should gain a level from that entry. That's currently how its is implemented from what I can see.

IanjhPhillips: I suppose it is possible that a user could create an entry that is both attached to a character as well as being of type 'dm' at creation. In this case the check here would be preventing such an entry from levelling that character. That's a change I can look into tomorrow.

tests/Feature/Http/Controllers/EntryControllerTest.php
pascwhale: Got it, thanks.

app/Observers/EntryObserver.php
```diff
@@ -28,11 +29,13 @@ public function created(Entry $entry)
      */
     public function updated(Entry $entry)
```
pascwhale, line 30: Just to be clear, this observer assumes that we cannot detach a DM Entry once we've attached it to a character, right? Correct me if I'm wrong, but if we were to remove a `character_id` from an entry, it would not substract the levels from the characters.

app/Observers/EntryObserver.php
```diff
@@ -28,11 +29,13 @@ public function created(Entry $entry)
...
     {
-        $character = $entry->character;
-        if ($entry->isDirty('levels')) {
-            $levelDelta = $entry->levels - $entry->getOriginal('levels');
-            $character->level += $levelDelta;
-            $character->save();
+        if ($entry->type != Entry::TYPE_DM || !is_null($entry->character)) {
+            $character = $entry->character;
+            if ($entry->isDirty('levels')) {
+                $levelDelta = $entry->levels - $entry->getOriginal('levels');
```
pascwhale, line 35: If the DM entry already had a level that was greater than zero but did not have a `character_id` attached to it, then the `levelDelta` would be null and no levels would be added to the newly attached character. I think this calls for a more robust set of rules in the `app/Http/Requests/EntryStoreRequest.php` file or more checks in this function. Let me know what you think.

app/Observers/EntryObserver.php
IanjhPhillips: Once an entry is added to a character, it should be deleted, not removed.

Upon deletion, the characters level will be updated to reflect the change

app/Observers/EntryObserver.php
IanjhPhillips: Good catch, I'll fix that right now.

pascwhale: I don't think that fixes the issue I was talking about. If a DM Entry got created with a level and no character, sending in a request that updates the level and attaches that DM Entry to a character would only add the difference in levels to that new character.

pascwhale (changes requested):

app/Observers/EntryObserver.php
IanjhPhillips: You're right. I rewrote the update method on the observer to catch this edge case.

pascwhale: Perfect, I think everything looks good. Nice work! I'm happy we have all cases sorted out.

pascwhale (approved):

PR #147, opened 2021-10-31
Character create, edit integration with backend

### LINKED ISSUE #100: Integrate character create with backend
## Description

Currently front end character create form is complete but it's not controlled and not integrated with backend, therefore we need to hook the UI up with backend.

## Possible Implementation

N/A

## Out of scope

N/A

N/A

### LINKED ISSUE #137: Integrate character edit and delete with backend
## Description

Currently, character management ui and backend character management apis are done, we need to hook the backend and frontend together.

## Possible Implementation

N/A
## Out of scope

N/A

N/A

willyyhuang:
Closes #100 
Closes #137


resources/js/Components/Form/CharacterCreateForm.tsx
```diff
@@ -55,82 +96,132 @@ const CharacterCreateForm = ({
                 {activeStep === 0 && (
                     <>
                         <Typography>
-                            Fill out the following fields with your
-                            character&apos;s details.
+                            Fill out the following fields with your character&apos;s
```
willyyhuang, line 99: really don't like hard coding strings like this, perhaps we can add a translation key system? will make code look much cleaner

m-triassi: Laravel has an in-built translation string system in the `resources/lang` folder, i could pass each of those files as universal props if you want, but i might end up being a lot of data

willyyhuang: if not, we can also use `i18next`, and manage it with json

willyyhuang: icebox issue i guess

pascwhale (changes requested):
The `resources/js/Components/Form/CharacterCreateForm.tsx` file exceeds 250 lines, which CodeClimate is flagging.
Aside from that, editing a character and confirming changes leaves the drawer open. It just goes back to the first step of the form.

routes/web.php
```diff
@@ -48,7 +48,10 @@
     Route::resource('entry-bulk', App\Http\Controllers\BulkEntryController::class)
         ->only(['store']);
 
-    Route::resource('character', App\Http\Controllers\CharacterController::class);
+    Route::delete('/character/{character?}', [App\Http\Controllers\CharacterController::class, 'destroy'])
```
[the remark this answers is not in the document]
m-triassi: Exactly, here we've made the character passed as a route parameter optional, meaning you can either make a DELETE request to `/character/1` or just `/character` so long as in the latter you pass a `characters` query string parameter :)

m-triassi:
> 
> 
> The `resources/js/Components/Form/CharacterCreateForm.tsx` file exceeds 250 lines, which CodeClimate is flagging. Aside from that, editing a character and confirming changes leaves the drawer open. It just goes back to the first step of the form.

Ah i see what's wrong, i can fix that really quick

willyyhuang:
> The `resources/js/Components/Form/CharacterCreateForm.tsx` file exceeds 250 lines, which CodeClimate is flagging. Aside from that, editing a character and confirming changes leaves the drawer open. It just goes back to the first step of the form.

we do plan to refactor the table component related to codeClimate issue, but that will be later, the focus should be on a function character module by the end of the sprint

pascwhale (approved):
Everything looks good. The fix for the drawer worked perfectly!

PR #152, opened 2021-11-04
Laptop resolution viewport overlap

willyyhuang:


pascwhale (approved):
Thanks Wilson, the headers are rendered a lot better on laptop screens.

PR #162, opened 2021-11-07
145 scope entry index

### LINKED ISSUE #114: Implement Update function on Entry Controller
## Description
Implement system specific logic for Update function on Entry Controller.
- update function should redirect to different index endpoint depending on the entry type.
- add a test case to confirm that this redirect occurs

Implement authorize logic on EntryUpdateRequest
- In authorize() on EntryUpdateRequest, if no character_id is present, then the entry must be of type DM and the user must be able to edit the entry in accordance with the existing policy. 
- Add a test case to check this.

### LINKED ISSUE #145: Scope Entry index to current user
## Description
Currently the Entry index() function returns all items for all characters. Instead items returned should be scoped either to a specific the character we care about, or alternatively to the currently auth'd user.

## Possible Implementation
Change the query in the EntryController index() function to accept a query parameter, "character_id", in the request. if it is null, the scope to the Auth::user().

JasonGadoury:
## Description

Closes #114 
Closes #145  

- Implements EntryController's index function to scope entry.index to specified user, or auth's user if user_id is null.
- Modifies EntryController's update function to redirect to entry or dm-entry index depending on Entry type
- Missing EntryPolicy mapping added to AuthServiceProvider


JasonGadoury:
Missing a test for 2nd condition in index function. Should do the trick for diff coverage...

app/Http/Controllers/EntryController.php
```diff
@@ -15,7 +16,13 @@ class EntryController extends Controller
      */
     public function index(Request $request)
     {
-        $entries = Entry::all();
+        if ($userId = $request->get('user_id')) {
+            $entries = Entry::where('user_id', $userId)->get();
+        } elseif (is_null($request->get('user_id'))) {
```
m-triassi, line 21: the request class has a function for this already, iirc; check the docs to be sure but i think its just:

```suggestion
        } elseif ($request->has('user_id')) {
```

app/Http/Controllers/EntryController.php
```diff
@@ -15,7 +16,13 @@ class EntryController extends Controller
...
+            $entries = Entry::where('user_id', Auth::user()->id)->get();
+        } else {
+            $entries = [];
+        }
```
m-triassi, line 25: i don't think this condition is reachable since `user_id` is either present or not

app/Http/Controllers/EntryController.php
m-triassi: Actually given my other comment about the else, you can probably just make this the `else` instead

m-triassi (changes requested):
Just some small tweaks, but overall seems fine!

app/Http/Controllers/EntryController.php
JasonGadoury: Done.

m-triassi (approved):
Excellent work! :) all green ✔️ s!

PR #163, opened 2021-11-07
refactor: #161 modularize cypress tests using Laravel integration

### LINKED ISSUE #161: Make use of Cypress commands integrating Laravel functions
## Description
[This package](https://github.com/laracasts/cypress) integrates Laravel-specific functions.

## Possible Implementation
The more interesting commands include logging in and creating model instances. This could particularly useful to create a test user more easily and keep each test within its scope. It will also allow tests to be more modular, since registering a user and logging them in can be handled without the use of the application's frontend components.

pascwhale:
## Description

Closes #161
Refactored tests to use method integrating Laravel API calls. Important functions to note include:
- cy.login
- cy.refreshDatabase
- cy.refreshRoute

These function help modularize tests and help each test focus on its true goal, instead of having extra lines to set up the test.

Tests also include ways to intercept request responses, making sure the response is received before moving on to the next step. This helps make sure a page is fully loaded before using up the next command's timeout.


RobertoNittolo (approved):
Good job! nice, refactor! I like how the tests are now separated.

PR #177, opened 2021-11-16
adventures league import bug fix

### LINKED ISSUE #172: Adventure's League Import Redirects to Error page for old characters
## Description
Higher level characters exported csv's cause the import to fail.

## Expected Behavior
Character is imported and user is presented with the character's entries.

## Actual Behavior
System redirects to an error page.

IanjhPhillips:
## Description
Closes #172 
Fixes bug which caused backend to throw an error for certain exported characters
- added checks for different entry types
- added functionality for importing items as well as entries

## How to test
pre-existing migration was modified so first:
`php artisan migrate:fresh`

then:
`php artisan tinker`
`$user = User::factory()->create()`
`Auth::login($user)`
`$character = AdventuresLeague::getCharacter(database_path("mocks/dante.csv"))`
observe that donte is imported, a character is created, and entries are attached.

observe the last entry and note down its id, ill refer to it as k, when you run the following command replace k with the id.
`$e = Entry::find(k)`
`$e->items`
observe the two items attached to this entry

### DIFF
--- a/app/Services/AdventuresLeagueAdapter.php
+++ b/app/Services/AdventuresLeagueAdapter.php
@@ -120,12 +120,12 @@ class AdventuresLeagueAdapter
                     'downtime' => (array_key_exists(8, $data[$i])) ? (float)$data[$i][8] : 0,
                     'type' => $type,
                 ];
-                $entry = Entry::factory()->create($entryData);
+                $entry = Entry::create($entryData);
                 $entry->save();
                 array_push($entries, $entry);
             } elseif ($isItemEntry) {
                 if (count($entries) == 0) {
-                    $e = Entry::factory()->create([
+                    $e = Entry::create([
                         'user_id' => Auth::id(),
                         'adventure_id' => null,
                         'campaign_id' => null,
--- a/app/Services/AdventuresLeagueAdapter.php
+++ b/app/Services/AdventuresLeagueAdapter.php
@@ -141,12 +141,15 @@ class AdventuresLeagueAdapter
                     $e->save();
                     array_push($entries, $e);
                 }
+
+                $rarity = ((array_key_exists(2, $data[$i])) ? (string) $data[$i][2] : "common");
                 $lastEntry = $entries[array_key_last($entries)];
+
                 $itemData = [
                     'entry_id' => $lastEntry->id,
                     'character_id' => $characterId,
                     'name' => (array_key_exists(1, $data[$i])) ? $data[$i][1] : "",
-                    //'rarity' => (array_key_exists(2, $data[$i])) ? $data[$i][2] : "common", //need enum
+                    'rarity' => in_array($rarity, Item::RARITY) ? $rarity : "uncommon", //need enum
                     'description' => (array_key_exists(6, $data[$i])) ? $data[$i][6] : "",
                     'author_id' => Auth::id(),
                     'tier' => 0,
--- a/app/Services/AdventuresLeagueAdapter.php
+++ b/app/Services/AdventuresLeagueAdapter.php
@@ -67,10 +67,13 @@ class AdventuresLeagueAdapter
         $character = new Character($characterData);
         $character->save();
 
+        //dd($character);
+
         // Get the entries from the file
         $entries = $this->getEntries($data, $character->id);
 
         $character->entries()->saveMany($entries);
+
         return $character;
     }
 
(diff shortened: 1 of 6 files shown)


app/Services/AdventuresLeagueAdapter.php
```diff
@@ -67,10 +67,13 @@ public function getCharacter($filePath): ?Character
         $character = new Character($characterData);
         $character->save();
 
+        //dd($character);
+
```
m-triassi, line 71: Lets remove this :)

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -83,33 +86,77 @@ public function getCharacter($filePath): ?Character
...
         for ($i = 4; $i < count($data); $i++) {
-            //Check if row is a magic item entry
             $isItemEntry = ($data[$i][0] == "MAGIC ITEM");
+            $isCharacterLogEntry = ($data[$i][0] == "CharacterLogEntry");
+            $isCampaignLogEntry = ($data[$i][0] == "CampaignLogEntry");
+            $isDmLogEntry = ($data[$i][0] == "DmLogEntry");
+            $isTradeLogEntry = ($data[$i][0] == "TradeLogEntry");
+            $isPurchaseLogEntry = ($data[$i][0] == "PurchaseLogEntry");
+
+            $isEntryLine = ($isCharacterLogEntry || $isDmLogEntry || $isCampaignLogEntry || $isTradeLogEntry || $isPurchaseLogEntry);
```
m-triassi, line 99: This could probably be extracted to a function? What do you think? or maybe it can just become
```php
$isEntryLine = in_array($data[i][0], ["CharacterLogEntry", "CampaignLogEntry", "DmLogEntry", "TradeLogEntry",  "PurchaseLogEntry"]
```

Additionally, we might need to reconsider if we will be turning `TradeLogEntry`s into entries in our system or not, or if we'll create dynamic entries from trades; currently i'm undecided :)

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -83,33 +86,77 @@ public function getCharacter($filePath): ?Character
...
 
-            if (!$isItemEntry) {
+            if ($isEntryLine) {
+                $type = "";
+                if ($isDmLogEntry) {
```
m-triassi, line 103: Oh i see, you need the `$isDmLogEntry` here, hmm ok, might be fine to ignore the above comment

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -83,33 +86,77 @@ public function getCharacter($filePath): ?Character
...
                 ];
-                array_push($entries, new Entry($entryData));
-            }
-            // Item importing to be refactored
-            /*else {
-                //create the item and attach it to the last entry
-                $latestEntry = end($entries);
-                if (is_null($latestEntry)) {
-                    continue;
+                $entry = Entry::factory()->create($entryData);
```
m-triassi, line 125: Not so sure we should be calling the factory here, thats usually reserved for testing. you can just do
```suggestion
                $entry = Entry::create($entryData);
```

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -83,33 +86,77 @@ public function getCharacter($filePath): ?Character
...
+                        'event_id' => null,
+                        'dungeon_master_id' => null,
+                        'levels' => 0,
+                        'date_played' => now(),
+                        'gp' => 0,
+                        'downtime' => 0,
+                        'type' => Entry::TYPE_GAME,
+                    ]);
+                    $e->save();
+                    array_push($entries, $e);
```
m-triassi, line 144: i'm having trouble understanding when this else is fallen into, when the line isn't an entry?

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -83,33 +86,77 @@ public function getCharacter($filePath): ?Character
...
-                //create the item and attach it to the last entry
-                $latestEntry = end($entries);
-                if (is_null($latestEntry)) {
-                    continue;
+                $entry = Entry::factory()->create($entryData);
+                $entry->save();
+                array_push($entries, $entry);
+            } else {
+                if (count($entries) == 0) {
+                    $e = Entry::factory()->create([
```
m-triassi, line 130: Same as above, no need to call `factory()`
```suggestion
                    $e = Entry::create([
```

database/migrations/2021_09_23_221026_create_items_table.php
```diff
@@ -19,7 +19,7 @@ public function up()
             $table->id();
             $table->foreignId('entry_id')->constrained();
             $table->foreignId('character_id')->constrained();
-            $table->string('name');
+            $table->text('name');
```
m-triassi, line 22: This is probably fine, but keep in mind we also have the option to keep is as a string, but instead increase the max character count, something like: 
```suggestion
            $table->string('name', 500);
```
should do the trick. Mostly because `text(...)` fields are stored as files, so they have a higher i/o cost to retrieve :)

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -83,33 +86,77 @@ public function getCharacter($filePath): ?Character
...
+                    array_push($entries, $e);
                 }
+                $lastEntry = $entries[array_key_last($entries)];
                 $itemData = [
-                    'name' => $data[$i][1],
-                    'rarity' => $data[$i][2],
+                    'entry_id' => $lastEntry->id,
+                    'character_id' => $characterId,
+                    'name' => (array_key_exists(1, $data[$i])) ? $data[$i][1] : "",
+                    //'rarity' => (array_key_exists(2, $data[$i])) ? $data[$i][2] : "common", //need enum
```
m-triassi, line 151: Ah yeah, this just needs to be added as a property on the `Item` model, since it's already set in the db, something like: 

`App\Models\Item.php`
```php
public const RARITY = ["common", "uncommon", "rare", "very_rare", "legendary"]
```

Then we'd just need some extra function that can constrain a string to one of those options

m-triassi (changes requested):
Just some small changes, honestly most of them are up to you, but there are some things that for sure need to be addressed :) Good overall though!!

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: Im using those individual bools in other combinations elsewhere though so i think keeping them atomic is best.

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: haha will do, good catch.

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: Sounds good ill change to this

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: hm now that im looking at it again it should be an else if. I'll need to give this another sit down.

database/migrations/2021_09_23_221026_create_items_table.php
IanjhPhillips: ah okay ill use string in that case!

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: I figured something like this, but when I coded it to take "common" as the default (as seen in the above commented code), it threw an error about truncating the field. A bit of research into it suggested it was because of the enum. That's about as far as i got there.

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -83,33 +84,80 @@ public function getCharacter($filePath): ?Character
...
+                        'dungeon_master_id' => null,
+                        'levels' => 0,
+                        'date_played' => now(),
+                        'gp' => 0,
+                        'downtime' => 0,
+                        'type' => Entry::TYPE_GAME,
+                    ]);
+                    $e->save();
+                    array_push($entries, $e);
                 }
```
m-triassi, line 143: Ahhh i see whats happening here now, if theres no recently created entry we make one to attach the item to! We might want to consider extracting some of these control flow pieces to separate functions during our next refactor of this. Either way looks good for now :)

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -83,33 +84,80 @@ public function getCharacter($filePath): ?Character
...
+                    'rarity' => in_array($rarity, Item::RARITY) ? $rarity : "uncommon", //need enum
+                    'description' => (array_key_exists(6, $data[$i])) ? $data[$i][6] : "",
+                    'author_id' => Auth::id(),
+                    'tier' => 0,
+                    'counted' => 0,
                 ];
-                $latestEntry->items()->save(new Item($itemData));
-            }*/
+                $item = new Item($itemData);
+                $item->save();
```
m-triassi, line 159: same here i think this will be a prime candidate for method extractions next time we refactor :)

m-triassi (approved):
Fantastic work :) The ALL site importer is getting more and more impressive!

I'll approve but i'm going to leave this unmerged to give @RobertoNittolo and @JasonGadoury a chance to take a look at this and understand it a little bit too

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: yea this is specifically to handle the case where the first line is an item with no entry defined above it.

PR #187, opened 2021-11-23
Add item creation when storing Entries

### LINKED ISSUE #183: Accomodate Items on Entry Store / Update
## Description

When creating a new entry or updating an existing one, the endpoint should support receiving an array of items details to be created as well.   

```
[
            'user_id' => 2,
            'adventure_id' => 120,
            'campaign_id' => null,
            'character_id' => 2,
            ...
            'downtime' => 10,
            'items' => [
               [
                   'name' => 'Longsword +2'
                   'rarity' => "uncommon"
                   ...
               ], 
               ...
           ]
]
```

## Possible Implementation

Extend `EntryStoreRequest` to also accept the above items array. Same for `EntryUpdateRequest`. Add step to create items in entry `store()` and `update()`

## Out of scope

n/a

Front-end should only have to make 1 request to create both the entry and items associated.

m-triassi:
## Description
Closes #183 

Adds the ability to specify and `items` array in the entry form on `EntryController@create()`. The items array should take a the format specified in #183. Each element in this `items` form array is an array with sub arrays. Each of these sub associative sub arrays specify properties of a particular item (These properties are found on the database, or the `$fillable` array on `Item`).

Installs a Laravel Action package, rather than using "handcrafted" actions, this provides a bit more structure and documentation! see [more information here](https://github.com/lorisleiva/laravel-actions).  

## How to Test
We're still faced with the same issue that doesn't allow for directly testing the endpoint due to sanctum + inertia, but you can test the added action instead, which captures most of the functionality of this PR .
1. Start tinker
2. use `Auth::loginUsingId(...)` to auth as some user in your `User` table
3. Create or query for an `Entry` saved to some variable `$e`
3. Call `CreateEntryItems::run($e, [['name' => 'some name...', 'rarity' => 'uncommon'], [...], ...])` fill the second parameter as you please.  
   - _note_: the fields `name`, and `rarity` are required fields, so unless checking error states, make sure to include those
4. Observe that the Entry has your items associated with it.

### DIFF
--- a/app/Http/Controllers/ItemController.php
+++ b/app/Http/Controllers/ItemController.php
@@ -77,7 +77,11 @@ class ItemController extends Controller
 
         $request->session()->flash('item.id', $item->id);
 
-        return redirect()->route('item.index');
+        if ($charId = $item->character_id) {
+            return redirect()->route('character.show', $charId);
+        } else {
+            return redirect()->route('dm-entry.index');
+        }
     }
 
     /**
@@ -87,8 +91,13 @@ class ItemController extends Controller
      */
     public function destroy(Request $request, Item $item)
     {
+        $charId = $item->character_id;
         $item->delete();
 
-        return redirect()->route('item.index');
+        if ($charId) {
+            return redirect()->route('character.show', $charId);
+        } else {
+            return redirect()->route('dm-entry.index');
+        }
     }
 }
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,4 +1,4 @@
 #!/bin/sh
-."$(dirname "$0")/_/husky.sh"
+./"$(dirname "$0")/_/husky.sh"
 
 npx --no-install commitlint --edit "$1"
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
-."$(dirname "$0")/_/husky.sh"
+./"$(dirname "$0")/_/husky.sh"
 
 npx lint-staged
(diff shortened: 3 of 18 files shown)


m-triassi:
We'll probably want to reflect the addition of actions as a design pattern in our class diagram. Justification for the addition is as follows: 
> To preserve and follow SOLID principals, namely the Single-responsibility paradigm, actions were added, this means that classes are much more specialized, making code more readable, and to_some_ extent self documenting.

app/Http/Controllers/ItemController.php
```diff
@@ -77,7 +77,11 @@ public function update(ItemUpdateRequest $request, Item $item)
 
         $request->session()->flash('item.id', $item->id);
 
-        return redirect()->route('item.index');
+        if ($charId = $item->character_id) {
```
IanjhPhillips, line 80: just so i'm sure i understand, this is checking if $item->character_id exists or rather isn't null?

IanjhPhillips (approved):
Ran the tinker test and confirmed that the action works.

All looks good to me! nicely done.

IanjhPhillips:
I may have to make some changes in my open PR #177 to accommodate the changes you made to the Item model and migrations, since I believe I'm manipulating the 'counted' field in my importer.

We should consider this when deciding which PR to close first.

app/Http/Controllers/ItemController.php
m-triassi: @IanjhPhillips That's exactly right, at the same time it sets a variable `$charId`; in loosely typed langauges, usually null is evaluated as a "falsey" value :)

m-triassi:
> I may have to make some changes in my open PR #177 to accommodate the changes you made to the Item model and migrations, since I believe I'm manipulating the 'counted' field in my importer.
> 
> We should consider this when deciding which PR to close first.

Ah, right that makes sense, it should be straight forward to fix in yours though, right? just remove any references to `counted`? if you'd like we can have a call about it? :)

PR #188, opened 2021-11-26
Add sonarcloud to CI/CD

### LINKED ISSUE #185: Add SonarQube to CI/CD pipeline
## Description

Add [SonarCloud](https://sonarcloud.io/) analysis as a set during the CI/CD Pipeline. This will supply missing metrics for our front-end application.

## Possible Implementation

Follow instructions on Sonar cloud Docs.

## Out of scope

n/a

Do not make SonarCloud analysis block PRs from merging.

RobertoNittolo:
## Description
Closes #185 
Add sonarcloud to the CI/CD pipeline
Add sonarcloud properties

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -137,3 +137,20 @@ jobs:
 
       - name: Run psalm
         run: ./vendor/bin/psalm
+
+  sonarcloud:
+    name: sonarcloud
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: SonarCloud Scan
+      uses: SonarSource/sonarcloud-github-action@v1.6
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
+
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,6 @@
+sonar.organization = DnD-Montreal_session-tome
+sonar.projectKey = dnd-montreal
+
+# relative paths to source directories. More details and properties are described
+# in https://sonarcloud.io/documentation/project-administration/narrowing-the-focus/
+sonar.sources = .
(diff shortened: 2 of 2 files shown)


RobertoNittolo:
A decision needs to be made if we want to use Sonarcloud automatic analysis or CI-based analysis. We can't have both options enabled. Currently, we have it set on automatic analysis. CI-based analysis may provide more control and configuration (if needed). The files added in this PR should allow for it to work using the CI. Also, it seems like I don't have permission on Sonarcloud to change some settings, such as quality gate and also the setting to decide if we want  CI-based analysis or automatic analysis

m-triassi:
> A decision needs to be made if we want to use Sonarcloud automatic analysis or CI-based analysis. We can't have both options enabled. Currently, we have it set on automatic analysis. CI-based analysis may provide more control and configuration (if needed). The files added in this PR should allow for it to work using the CI. Also, it seems like I don't have permission on Sonarcloud to change some settings, such as quality gate and also the setting to decide if we want CI-based analysis or automatic analysis

I'd say we should go for CI-based analysis, since that way we can set the sonar cloud step to not make the pipeline fail, considering we're primarily interested in the metric it can provide.

Though for the CI-Based analysis, should you be passing the coverage reports to SonarCloud?

RobertoNittolo:
> > A decision needs to be made if we want to use Sonarcloud automatic analysis or CI-based analysis. We can't have both options enabled. Currently, we have it set on automatic analysis. CI-based analysis may provide more control and configuration (if needed). The files added in this PR should allow for it to work using the CI. Also, it seems like I don't have permission on Sonarcloud to change some settings, such as quality gate and also the setting to decide if we want CI-based analysis or automatic analysis
> 
> I'd say we should go for CI-based analysis, since that way we can set the sonar cloud step to not make the pipeline fail, considering we're primarily interested in the metric it can provide.
> 
> Though for the CI-Based analysis, should you be passing the coverage reports to SonarCloud?

We could pass in the coverage reports however, we have code climate which is taking care of all that. I was under the impression that we would be using sonarcloud strictly to cover for the unaccounted metrics needed and to continue using code climate the way we have been using it.

m-triassi (approved):
> We could pass in the coverage reports however, we have code climate which is taking care of all that. I was under the impression that we would be using sonarcloud strictly to cover for the unaccounted metrics needed and to continue using code climate the way we have been using it.

Ah yeah that's true, i almost wonder if we should not require the SonarCloud check to have a passing pipeline? its a tough call since there's a lot of shared functionality between SonarCloud and CodeClimate 🤔 Lets add this for now and if at some point having both becomes too cumbersome, we'll delete one or the other. 

Approved!

PR #191, opened 2021-11-28
190 item import fix

### LINKED ISSUE #190: Remove deprecated "counted" field from AL item import
## Description

Al importer uses deprecated "counted" field when creating imported items. Should be removed.

IanjhPhillips:
## Description
closes #190 
- removes deprecated field manipulation
- refactors entry data by extracting to method call

### DIFF
--- a/app/Services/AdventuresLeagueAdapter.php
+++ b/app/Services/AdventuresLeagueAdapter.php
@@ -64,8 +64,7 @@ class AdventuresLeagueAdapter
         ];
 
         // Hydrate Model
-        $character = new Character($characterData);
-        $character->save();
+        $character = Character::create($characterData);
 
         // Get the entries from the file
         $entries = $this->getEntries($data, $character->id);
@@ -105,39 +104,14 @@ class AdventuresLeagueAdapter
                 } else {
                     $type = Entry::TYPE_DOWNTIME;
                 }
-                $entryData = [
-                    'user_id' => Auth::id(),
-                    'adventure_id' => null,
-                    'campaign_id' => null,
-                    'character_id' => $characterId,
-                    'event_id' => null,
-                    'dungeon_master_id' => null,
-                    'dungeon_master' => (array_key_exists(12, $data[$i])) ? (float)$data[$i][12] : "",
-                    'location' => (array_key_exists(11, $data[$i])) ? (float)$data[$i][11] : "",
-                    'levels' => 0,
-                    'date_played' => ($data[$i][3] == "") ? now() : $data[$i][3],
-                    'gp' => (array_key_exists(7, $data[$i])) ? (float)$data[$i][7] : 0,
-                    'downtime' => (array_key_exists(8, $data[$i])) ? (float)$data[$i][8] : 0,
-                    'type' => $type,
-                ];
+                $entryData = $this->getEntryData($characterId, $data[$i], $type);
                 $entry = Entry::create($entryData);
                 $entry->save();
                 array_push($entries, $entry);
             } elseif ($isItemEntry) {
                 if (count($entries) == 0) {
-                    $e = Entry::create([
-                        'user_id' => Auth::id(),
-                        'adventure_id' => null,
-                        'campaign_id' => null,
-                        'character_id' => $characterId,
-                        'event_id' => null,
-                        'dungeon_master_id' => null,
-                        'levels' => 0,
-                        'date_played' => now(),
-                        'gp' => 0,
-                        'downtime' => 0,
-                        'type' => Entry::TYPE_GAME,
-                    ]);
+                    $entryData = $this->getEntryData($characterId, null, null);
+                    $e = Entry::create($entryData);
                     $e->save();
                     array_push($entries, $e);
                 }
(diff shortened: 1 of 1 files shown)


m-triassi:
Just a comment, before i do any reviewing, for your conventional commit label you can use `refactor: #190 reduce...` instead of `chore: #190 reduce...` in the future ;) especially since that was definitely a method extraction refactor

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -153,12 +127,43 @@ private function getEntries($data, $characterId)
...
+            return array_merge($entryData, $defaultEntryData);
+        } else {
+            $populatedEntryData = [
+                'dungeon_master' => (array_key_exists(12, $data)) ? (float)$data[12] : "",
+                'location' => (array_key_exists(11, $data)) ? (float)$data[11] : "",
+                'date_played' => ($data[3] == "") ? now() : $data[3],
+                'gp' => (array_key_exists(7, $data)) ? (float)$data[7] : 0,
+                'downtime' => (array_key_exists(8, $data)) ? (float)$data[8] : 0,
+                'type' => $type,
+            ];
```
m-triassi, line 165: might be more readable to wrap `$data` in a collection and then use the [`get()` method](https://laravel.com/docs/8.x/collections#method-get), but i'd say thats mostly a matter of opinion
```suggestion
            $data = collect($data)
            $populatedEntryData = [
                'dungeon_master' => (float) $data->get(12),
                'location' => (float) $this->get(11),
                'date_played' => Carbon::parse($data->get(3)),
                'gp' => (float) $data->get(7, 0),
                'downtime' => (float) $data->get(8, 0),
                'type' => $type,
            ];
```

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -153,12 +127,43 @@ private function getEntries($data, $characterId)
...
+                'date_played' => now(),
+                'gp' => 0,
+                'downtime' => 0,
+                'type' => Entry::TYPE_GAME,
+            ];
+            return array_merge($entryData, $defaultEntryData);
+        } else {
+            $populatedEntryData = [
+                'dungeon_master' => (array_key_exists(12, $data)) ? (float)$data[12] : "",
+                'location' => (array_key_exists(11, $data)) ? (float)$data[11] : "",
```
m-triassi, line 160: wait are these supposed to be cast to a float? dungeon master and location are supposed to be a strings describing either item

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -153,12 +127,43 @@ private function getEntries($data, $characterId)
...
                 ];
-                $item = new Item($itemData);
-                $item->save();
+                Item::create($itemData);
             }
         }
         return $entries;
     }
+
+    private function getEntryData($characterId, $data, $type)
```
m-triassi, line 137: since you're checking data is null later in this function, we should probably define it as default null, same with type
```suggestion
    private function getEntryData($characterId, $data = null, $type = null)
```

app/Services/AdventuresLeagueAdapter.php
```diff
@@ -105,39 +104,14 @@ private function getEntries($data, $characterId)
...
-                        'character_id' => $characterId,
-                        'event_id' => null,
-                        'dungeon_master_id' => null,
-                        'levels' => 0,
-                        'date_played' => now(),
-                        'gp' => 0,
-                        'downtime' => 0,
-                        'type' => Entry::TYPE_GAME,
-                    ]);
+                    $entryData = $this->getEntryData($characterId, null, null);
```
m-triassi, line 113: if you set defaults in the evoked function signature this can just be 
```suggestion
                    $entryData = $this->getEntryData($characterId);
```

m-triassi (commented):
Honestly i can probably approve here, imo, but let me know what you think of my suggestions first :)

RobertoNittolo (commented):
Good Job, Great refactor! Pulled it locally everything looked good, just wasn't sure how to test it.

RobertoNittolo:
> Good Job, Great refactor! Pulled it locally everything looked good, just wasn't sure how to test it.

I was able to find some previous instructions on how to test this from a previous PR. 

## How to test
pre-existing migration was modified so first:
`php artisan migrate:fresh`

then:
`php artisan tinker`
`$user = User::factory()->create()`
`Auth::login($user)`
`$character = AdventuresLeague::getCharacter(database_path("mocks/dante.csv"))`
observe that donte is imported, a character is created, and entries are attached.

observe the last entry and note down its id, ill refer to it as k, when you run the following command replace k with the id.
`$e = Entry::find(k)`
`$e->items`
observe the two items attached to this entry

app/Services/AdventuresLeagueAdapter.php
pascwhale: I never knew PHP had that feature! Thank you for mentioning it.

app/Services/AdventuresLeagueAdapter.php
pascwhale: Is dungeon master the ID of the dungeon master or a string to refer to that person (maybe not their actual name)? I also remember the D&D MTL talking about possibly not having accounts for someone acting as the DM.

app/Services/AdventuresLeagueAdapter.php
pascwhale: @m-triassi Can you tell me what happens if the index does not exist for a given collection through the get method? I couldn't find documentation online.

pascwhale (commented):
I really liked the refactored code.

What I'd like to add on top of the other suggestions would be to avoid using indices to get specific columns values. If we know the default column names for this type of file using those names to get the column values might be safer and would make our code more readable.

I do not have enough experience using those CSV files though, and I don't know if the format might change. Maybe someone else would know.

app/Services/AdventuresLeagueAdapter.php
m-triassi: @pascwhale haha you didn't read closely enough, first line of the get methods docs:
> The get method returns the item at a given key. **If the key does not exist, null is returned**

app/Services/AdventuresLeagueAdapter.php
m-triassi: > Is dungeon master the ID of the dungeon master or a string to refer to that person

In this case its going to be their name as a string

> I also remember the D&D MTL talking about possibly not having accounts for someone acting as the DM.

Right, that's why in the database we have 2 fields, if the user is not present in our system then they can enter the string representation of the DMs name

app/Services/AdventuresLeagueAdapter.php
m-triassi: no problem :)

m-triassi:
> What I'd like to add on top of the other suggestions would be to avoid using indices to get specific columns values

@pascwhale I can answer this for Ian, sadly, we are forces to use a specific index given the state of the CSV that the ALL site produces, it's quite _ahem_ non-standard...

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: not sure why these are being cast to float here, as you said they should be strings

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: great suggestion, added!

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: done!

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: fixed.

app/Services/AdventuresLeagueAdapter.php
IanjhPhillips: I like this implementation, its a lot cleaner. I added it. Thanks!

m-triassi (approved):
Excellent stuff! :) No more comments on my end, lets merge it 🚀

PR #196, opened 2021-11-28
Dm entries create edit form

### LINKED ISSUE #176: DM entries create/edit form
image

image

bcobo341:
Closes #176 

Add DM Entry create/edit form


resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
+import {Link, useForm, usePage} from '@inertiajs/inertia-react'
```
willyyhuang, line 1: we have a new `Link` component in atom folder, use that otherwise it will have no styling

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+    Box,
+    Button,
+    Grid,
+    Step,
+    StepLabel,
+    Stepper,
+    TextField,
+    Typography,
+} from '@mui/material'
+import Alert from '@mui/material/Alert'
```
willyyhuang, line 15: merge this with the import from `@mui/material` please, same for Chip, FormControl and all other components that can be imported from a single module

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+    min-width: 25vw;
+`
+
+const DmEntryCreateForm = ({
+    type,
+    onCloseDrawer = () => {},
+    editData,
+    editId = 0,
+}: DmEntryCreateFormPropType) => {
+    const {auth, adventures} = usePage<UsePageType>().props
```
willyyhuang, line 84: don't use `usePage`, pass down from parent component please,
`usePage` is only to use for global component that exists in every single page like `Authentication`

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+        type === 'Create'
+            ? DM_ENTRY_CREATE_FORM_INITIAL_VALUE
+            : {
+                  adventure_id: editData?.adventure || 0,
+                  length: editData?.length || 0,
+                  location: editData?.location || '',
+                  date_played: editData?.date_played || new Date().toDateString(),
+                  notes: editData?.notes || '',
+                  items: editData?.items || [],
+                  type: 'dm',
```
willyyhuang, line 113: this property will only have this value?

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+                  user_id: user?.id,
+              }
+
+    const {data, setData, errors, clearErrors, post, put} = useForm(
+        DM_ENTRY_FORM_INITIAL_VALUE,
+    )
+    const [activeStep, setActiveStep] = useState<number>(0)
+    const [item_data, setItemData] = useState<ItemData>({
```
willyyhuang, line 121: dont think you should save `item_data` in state, they will be change, save it in `items` instead

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+        name: '',
+        description: '',
+        rarity: '',
+        tier: 1,
+    })
+    const [items, setItems] = useState<ItemData[]>([])
+    const [isItemsVisible, setIsItemsVisible] = useState<boolean>(false)
+    const [error, setError] = useState<boolean>(false)
```
willyyhuang, line 129: error if i remember correctly should be server side response, hence it should be in props

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+                                                setItemData(
+                                                    ITEM_CREATE_FORM_INITIAL_VALUE,
+                                                )
+                                            } else {
+                                                setError(true)
+                                            }
+                                        }}>
+                                        Add Item
+                                    </Button>
+                                ) : null}
```
willyyhuang, line 416: don't use ternary operator when the other condition is null, rahter do `type === 'Create' && ()`

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+                                        }
+                                    }}>
+                                    {type === 'Create' ? 'Create' : 'Save'}
+                                </Button>
+                            </Grid>
+                        </Grid>
+                    </>
+                )}
+            </StyledFooter>
+            {isItemsVisible ? (
```
willyyhuang, line 450: same here

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,479 @@
...
+                <StyledItemsFooter container spacing={1}>
+                    <StyledGrid item xs={12}>
+                        <Typography variant='body2'>New Magic Items</Typography>
+                    </StyledGrid>
+                    {error ? (
```
willyyhuang, line 455: same here

willyyhuang (changes requested):

willyyhuang:
also merge main please

resources/js/Components/Form/DmEntryCreateForm.tsx
bcobo341: The type will only have the value 'dm' inside the DmEntryCreate form as you can only create it from here and needs to be specified when making a request.

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -24,6 +24,8 @@ import {UsePageType} from 'Types/global'
 import {ItemData} from 'Types/item-data'
 import route from 'ziggy-js'
 
+import Link from '../Atom/Link'
```
willyyhuang, line 27: you can just import from 'Components'

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -353,9 +359,10 @@ const DmEntryCreateForm = ({
                         <Grid container spacing={2} />
                         <Grid item md={type === 'Edit' ? 4 : 2} xs={6}>
                             {type === 'Create' ? (
-                                <Link href={route('character.index')}>
-                                    <Button fullWidth>Cancel</Button>
-                                </Link>
+                                <Link
+                                    href={route('character.index')}
```
willyyhuang, line 363: dm entry route is in ziggy now i believe, ask annie

resources/js/Components/Form/DmEntryCreateForm.tsx
willyyhuang: ok

resources/js/Components/Form/DmEntryCreateForm.tsx
bcobo341: right, I'll correct that

m-triassi:
I might be able to help out with the coverage here, looks like some PHP isn't covered :)

m-triassi:
Looks like i was only able to get you 2% diff coverage 😅

willyyhuang (changes requested):

resources/js/Pages/Entry/Create/DmEntryCreate.tsx
```diff
@@ -0,0 +1,24 @@
...
+import {ThemeProvider} from '@mui/material/styles'
+import {DmEntryCreateForm} from 'Components'
+import {ApplicationLayout} from 'Layouts'
+import React from 'react'
+import {UsePageType} from 'Types/global'
+import {getFontTheme} from 'Utils'
+
+const DmEntryCreate = () => {
+    const theme = getFontTheme('Form', 16)
+    const {auth, adventures} = usePage<UsePageType>().props
```
willyyhuang, line 11: don't use this, get it from props

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,359 @@
...
+                  user_id,
+              }
+
+    const {data, setData, errors, clearErrors, post, put} = useForm(
+        DM_ENTRY_FORM_INITIAL_VALUE,
+    )
+    const [activeStep, setActiveStep] = useState<number>(0)
+    const [items, setItems] = useState<ItemData[]>([])
+    const [isItemsVisible, setIsItemsVisible] = useState<boolean>(false)
+    const [isDuplicate, setIsDuplicate] = useState<boolean>(false)
```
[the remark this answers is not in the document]
bcobo341: to check for duplicates as you don't want to be adding 2 items with the same name for the same entry

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,359 @@
...
+                  type: 'dm',
+                  user_id,
+              }
+
+    const {data, setData, errors, clearErrors, post, put} = useForm(
+        DM_ENTRY_FORM_INITIAL_VALUE,
+    )
+    const [activeStep, setActiveStep] = useState<number>(0)
+    const [items, setItems] = useState<ItemData[]>([])
+    const [isItemsVisible, setIsItemsVisible] = useState<boolean>(false)
```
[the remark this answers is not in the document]
bcobo341: I don't think addItemForm is the correct choice of wording. This state is used to hide the chip items when needed.

resources/js/Pages/Entry/Create/DmEntryCreate.tsx
bcobo341: how? the component isn't accessible from the ApplicationLayout

resources/js/Components/Form/DmEntryCreateForm.tsx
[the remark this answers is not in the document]
willyyhuang: is this a condition from the entry logic? what if someone gets 2 items of the same name

resources/js/Components/Form/DmEntryCreateForm.tsx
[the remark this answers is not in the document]
bcobo341: This state was removed completely

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,359 @@
...
+    const [isDuplicate, setIsDuplicate] = useState<boolean>(false)
+
+    const handleDeleteItem = (chipToDelete: ItemData) => {
+        setItems((chips) => chips.filter((chip) => chip.name !== chipToDelete.name))
```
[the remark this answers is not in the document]
bcobo341: items state is now removed

resources/js/Pages/Entry/Create/DmEntryCreate.tsx
willyyhuang: `const DmEntryCreate = ({adventures, auth}) => `

will get you what you want

willyyhuang: obviously you need to type them

resources/js/Components/Form/ItemCreateForm.tsx
```diff
@@ -0,0 +1,224 @@
...
+                    <StyledGrid item xs={12} md={type === 'Edit' ? 12 : 5}>
+                        <TextField
+                            margin='normal'
+                            fullWidth
+                            id='name'
+                            label='Name'
+                            name='Name'
+                            value={data.name}
+                            onChange={(e) =>
+                                setData({
```
willyyhuang, line 90: you can just use 2 args instead of this
setData('name', value)

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+import ItemCreateForm from './ItemCreateForm'
+
+type DmEntryCreateFormPropType = {
+    type: 'Edit' | 'Create'
+    onCloseDrawer?: () => void
+    editData?: EntriesData
+    editId?: number
+    user_id?: number
+    adventures?: any
```
willyyhuang, line 32: you can import the type for this

willyyhuang:
can you pull the latest main branch and connect the form to the index page

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+import ItemCreateForm from './ItemCreateForm'
+
+type DmEntryCreateFormPropType = {
+    type: 'Edit' | 'Create'
+    onCloseDrawer?: () => void
+    editData?: EntriesData
+    editId?: number
+    user_id?: number
+    adventures?: adventureType[]
```
willyyhuang, line 33: i think adventures should be required no?

m-triassi: Well we left them nullable because when we import from DnDBeyond we have no adventure data and the user has to fill that out later, so it sadly needs to be nullable

willyyhuang: make sense, can we have an initial value for it tho, if it's nullable just return `[]` perhaps?

m-triassi (approved):
I'm going to make an executive decision here and say this is good to be merged now.

PR #200, opened 2021-11-29
184 dm entry reward

### LINKED ISSUE #184: Add DM Entry reward choice logic
## Description

When a DM assigns an entry to a character they must choose whether they want to gain a level or an item from the assignment. there should be a field that that determines what action to take

## Possible Implementation

A if statement can be added to the entry store / update functions that would either change the level of the entry to 0 and attach the selected item, or toss away the item and set the level of the entry to 1, thus letting the observer take over.

## Out of scope

n/a

A special field should be added to the form request on the front-end. this will also entail changing the Store and Update request classes.

JasonGadoury:
### Description

Closes #184 

Adds EntryController logic to process store and update requests for DM Entries which contain a reward choice. The possible choices include advancement, magic_item and campaign_reward. The store and update requests must modify the items or levels attributes in the entry based on the reward choice.

### DIFF
--- a/app/Http/Controllers/EntryController.php
+++ b/app/Http/Controllers/EntryController.php
@@ -7,6 +7,7 @@ use App\Http\Requests\EntryStoreRequest;
 use App\Http\Requests\EntryUpdateRequest;
 use App\Models\Entry;
 use Illuminate\Http\Request;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Auth;
 
 class EntryController extends Controller
@@ -44,10 +45,19 @@ class EntryController extends Controller
         $entryData = collect($request->validated())->except('items');
         $itemData = collect($request->validated())->only('items');
 
+        if ($itemData->has('items')) {
+            $itemData = $itemData['items'];
+        }
+
+        if ($itemData instanceof Collection) {
+            $itemData = $itemData->toArray();
+        }
+
         //possibly implement as function for reusability
         if ($entryData->get('choice') == 'advancement') {
             // advancement: increment character's level
             $entryData['levels'] = 1;
+            $itemData = [];
         } elseif ($entryData->get('choice') == 'magic_item') {
             // magic_item: attach item to character of choice, and set entry's levels to 0
             $itemData = [$itemData[0] ?? []];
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/EntryController.php
```diff
@@ -43,10 +44,33 @@ public function store(EntryStoreRequest $request)
...
+        } elseif ($entryData->get('choice') == 'magic_item') {
+            // magic_item: attach item to character of choice, and set entry's levels to 0
+            $itemData = [$itemData[0] ?? []];
+            $entryData['levels'] = 0;
+        } elseif ($entryData->get('choice') == 'campaign_reward') {
+            // campaign reward: set levels = 0, no item(s), should contain custom note
+            $itemData = [];
+            $entryData['levels'] = 0;
+        }
+
```
m-triassi, line 70: The more i stare at this the more i definitely think we're gonna need to refactor this into a seperate function of some sort, but it'll do for now

app/Http/Controllers/EntryController.php
```diff
@@ -85,8 +109,32 @@ public function update(EntryUpdateRequest $request, Entry $entry)
...
+            $itemData = [];
+        } elseif ($entryData->get('choice') == 'magic_item') {
+            // magic_item: attach item to character of choice, and set entry's levels to 0
+            $itemData = [$itemData[0] ?? []];
+            $entryData['levels'] = 0;
+        } elseif ($entryData->get('choice') == 'campaign_reward') {
+            // campaign reward: set levels = 0, no item(s), should contain custom note
+            $itemData = [];
+            $entryData['levels'] = 0;
+        }
```
m-triassi, line 134: same as above here

tests/Feature/Http/Controllers/EntryControllerTest.php
```diff
@@ -270,6 +271,112 @@ public function store_saves_entry_with_items()
...
+            'event_id' => $event->id,
+            'dungeon_master_id' => $dungeon_master_user->id,
+            'dungeon_master' => $dungeon_master,
+            'date_played' => $date_played,
+            'location' => $location,
+            'type' => $type,
+            'levels' => $levels,
+            'gp' => $gp,
+            'choice' => $choice_advancement,
+
```
m-triassi, line 316: we can probably extract this to its own array since its the same in all three post requests, lets do that in a refactor of this feature next sprint

m-triassi (approved):
Definitely something to improve here, but given its the end of the sprint, lets just ship this and refactor later since it at least works

app/Http/Controllers/EntryController.php
JasonGadoury: Agreed, can be used for both update and store methods

Issue #208, opened 2021-12-05
Generate User Rating Report


m-triassi:
## Description

There should be a general report that can be generated by site super admins that compiles a list of all players / DMs that have played at DnD MTL events over a given date interval. The report should include a users overall rating / a breakdown of their strengths and weaknesses.

## Possible Implementation

Laravel Backpack does seem to [support table level operations](https://backpackforlaravel.com/docs/4.1/crud-operations#standard-operations), this might be a good way to achieve this. Otherwise a [single action controller](https://laravelactions.com/2.x/as-controller.html) might also be a viable solution for this.

## Requirements / Restrictions

There must be the ability to filter out non-event ratings from the result set, seeing as home games are mostly irrelevant for the purpose of statistics tracking withing DnD MTL.

m-triassi:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Session-Tome-613674a263226c0015d44c06/issues/DnD-Montreal/session-tome/208?planning-poker) with ZenHub @bcobo341 @IanjhPhillips @JasonGadoury @RobertoNittolo

PR #220, opened 2021-12-27
Rating Categories

### LINKED ISSUE #205: Modify Rating Model
## Description

Rating model should be modified to accommodate Rating System requirements as outlined by stakeholders. Ratings should be able to be of a predefined set of types. (such as "helpful", "creative, "accommodating", "negative", etc...)
More details to be added upon stakeholder's response. 

## Possible Implementation

Depending on number of rating types per rating instance, can either have a simple 'type' column added on to the model with a predefined constant set of types, or, if multiple types per rating are required, consider using a [bitmask](https://aaronfrancis.com/2021/bitmasking-in-laravel-and-mysql) implementation to handle the typing.

IanjhPhillips:
## Description
Closes #205 
Adds the 'categories' field to the Ratings model, which is an integer that tracks which categories of rating a given rating instance has by making use of bit manipulation and bitmasks.

Also adds two query scopes which can be used to filter by rating category, they are
- Rating::withCategories()
- Rating::withoutCategories()

An example of a query using one of the above scopes:
`$friendlyHelpfulRating = Rating::withCategories([
            Rating::FRIENDLY_BITMASK,
            Rating::HELPFUL_BITMASK,
        ])->first();`

Also removes the score field from Ratings model, which is no longer needed due to the updated design of the rating system.

## How to test
open tinker and create a Rating instance, call it `$rating`

Observe the categories field, if it is non zero run the following command, if it is zero it is not a problem, simple skip this command, or run it if you'd like to see an empty list.
`$rating->labels`
Observe the list of rating categories

Next, add a category, selecting from one of the five bitmask constants declared on the Model. 
For your convenience, here is a list of the categories: [CREATIVE, FLEXIBLE, FRIENDLY, HELPFUL, PREPARED]
(example command below)
`$rating->addCategory(Rating::PREPARED_BITMASK)`

Now, check that the category has been added by re running the label list command from earlier.
`$rating->labels`

Next, remove a category, again selecting a category that fits your case.
`$rating->removeCategory(Rating::PREPARED_BITMASK)`

Finally, check that the category has been removed.
`$rating->labels`

### DIFF
--- a/app/Models/Rating.php
+++ b/app/Models/Rating.php
@@ -119,4 +119,25 @@ class Rating extends Model
 
         return $bitmasks->only($bits);
     }
+
+    public function scopeHasCategories($query, $bits)
+    {
+        $this->queryCategoriesBitmask($query, $bits, $containsCategories = true);
+    }
+
+    public function scopeMissingCategories($query, $bits)
+    {
+        $this->queryCategoriesBitmask($query, $bits, $containsCategories = true);
+    }
+
+    protected function queryCategoriesBitmask($query, $bits, $containsCategories)
+    {
+        $bits = array_sum(Arr::wrap($bits));
+
+        $queryOperator = $containsCategories ? '=' : '!=';
+
+        //build the query
+        // (categories & bits) <operator> bits
+        $query->whereRaw("categories & $bits $queryOperator $bits");
+    }
 }
--- a/app/Models/Rating.php
+++ b/app/Models/Rating.php
@@ -5,6 +5,7 @@ namespace App\Models;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Arr;
 
 class Rating extends Model
 {
@@ -127,17 +128,16 @@ class Rating extends Model
 
     public function scopeMissingCategories($query, $bits)
     {
-        $this->queryCategoriesBitmask($query, $bits, $containsCategories = true);
+        $this->queryCategoriesBitmask($query, $bits, $containsCategories = false);
     }
 
     protected function queryCategoriesBitmask($query, $bits, $containsCategories)
     {
         $bits = array_sum(Arr::wrap($bits));
-
         $queryOperator = $containsCategories ? '=' : '!=';
 
         //build the query
-        // (categories & bits) <operator> bits
+        //(categories & bits) <operator> bits
         $query->whereRaw("categories & $bits $queryOperator $bits");
     }
 }
--- a/app/Models/Rating.php
+++ b/app/Models/Rating.php
@@ -19,6 +19,7 @@ class Rating extends Model
         'user_id',
         'author_id',
         'score',
+        'categories',
     ];
 
     /**
@@ -31,7 +32,14 @@ class Rating extends Model
         'entry_id' => 'integer',
         'user_id' => 'integer',
         'author_id' => 'integer',
+        'categories' => 'integer',
     ];
+
+    public const CREATIVE_BITMASK = 1 << 0;    // 00001
+    public const FLEXIBLE_BITMASK = 1 << 1;    // 00010
+    public const FRIENDLY_BITMASK = 1 << 2;    // 00100
+    public const HELPFUL_BITMASK = 1 << 3;     // 01000
+    public const PREPARED_BITMASK = 1 << 4;    // 10000
 
 
     public function entry()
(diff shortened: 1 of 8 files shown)


app/Models/Rating.php
```diff
@@ -48,4 +62,92 @@ public function author()
...
+        ]);
+
+        foreach ($bitmasks as $mask) {
+            if ($this->hasCategory($mask)) {
+                $labels->push($this->getCategoryLabelFromMask($mask));
+            }
+        }
+
+        return $labels;
+    }
```
m-triassi, line 117: lets turn this into an [accessor](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor) instead :) Something like `getLabelsAttribute()`

We can also save on evaluating all that control flow in [getCategoryLabelFromMask()](https://github.com/DnD-Montreal/session-tome/blob/9b95bc73c4487d94896d27342a6d92c85921facd/app/Models/Rating.php#L124) if we make the collection look like this: 

```php
$bitmasks = collect([
   self::CREATIVE_BITMASK => self::CREATIVE_LABEL,
   self::FLEXIBLE_BITMASK => self::FLEXIBLE_LABEL,
   self::FRIENDLY_BITMASK => self::FRIENDLY_LABEL,
   self::HELPFUL_BITMASK => self::HELPFUL_LABEL,
   self::PREPARED_BITMASK => self::PREPARED_LABEL,
])
```
 
Decomposing the `$this->categories` into an array of its powers of two and then calling `$bitmasks->only(...)`

Decomposition will look something like this: 
```php
$bit = 1;
$select = []
while ($bit <= $this->categories) {
   if ($bit & $this->categories) {
      $select[] = $bit
   }
   $bit = $bit << 1
}
```

app/Models/Rating.php
```diff
@@ -48,4 +62,92 @@ public function author()
...
+
+            case self::PREPARED_BITMASK:
+                $label = self::PREPARED_LABEL;
+                break;
+
+            default:
+                $label = "";
+        }
+        return $label;
+    }
```
m-triassi, line 152: Lets also add some query scopes since we'll need to be able to filter on this in #208, The article you referenced [lays out how to do that](https://aaronfrancis.com/2021/bitmasking-in-laravel-and-mysql#bitmasks-in-mysql) pretty nicely

If you need more info in general about query scopes, [the docs also have some good info](https://laravel.com/docs/8.x/eloquent#query-scopes)

app/Models/Rating.php
```diff
@@ -48,4 +62,92 @@ public function author()
...
+
+        return $labels;
+    }
+
+    /**
+     * Get category label corresponding to a bitmask
+     * @param $mask integer The bitmask
+     * @return string The Label
+     */
+    private function getCategoryLabelFromMask($mask)
```
m-triassi, line 124: theoretically this whole method can go away if you add the changes on Line 99

m-triassi (changes requested):
This is a fantastic start and i think if we address a few things its going to be absolutely perfect :)

pascwhale (approved):
Everything looks good to me.

app/Models/Rating.php
IanjhPhillips: done! Its now an accessor so the labels can be found via `$rating->labels` which returns a collection of `bitmask => label`

app/Models/Rating.php
IanjhPhillips: removed it

app/Models/Rating.php
IanjhPhillips: all done!

m-triassi (approved):
WOWOWOW! Fantastic!! :)

app/Models/Rating.php
```diff
@@ -48,4 +63,81 @@ public function author()
...
+        $this->categories &= ~$bitMask;
+    }
+
+    /**
+     * Returns category labels. Not recommended for large batches of requests, instead use hasCategory with bitmask
+     * @returns Collection of strings corresponding to the rating categories of this entity
+     */
+    public function getLabelsAttribute()
+    {
+        $labels = collect();
```
m-triassi, line 102: un-unused variable! Just gonna clean up some of your code smells really quick
```suggestion
```

tests/Unit/Models/RatingTest.php
```diff
@@ -40,4 +40,92 @@ public function can_belong_to_author()
...
+     */
+    public function can_scope_queries_to_categories()
+    {
+        $categories1 = Rating::FLEXIBLE_BITMASK + Rating::FRIENDLY_BITMASK;
+        $categories2 = Rating::HELPFUL_BITMASK + Rating::FRIENDLY_BITMASK;
+        $categories3 = Rating::PREPARED_BITMASK + Rating::HELPFUL_BITMASK;
+
+        $rating1 = Rating::factory()->create(['categories' => $categories1]);
+        $rating2 = Rating::factory()->create(['categories' => $categories2]);
+        $rating3 = Rating::factory()->create(['categories' => $categories3]);
```
m-triassi, line 117: We don't really need to assign these to variable if they wont be used

```suggestion
        Rating::factory()->create(['categories' => $categories1]);
        Rating::factory()->create(['categories' => $categories2]);
        Rating::factory()->create(['categories' => $categories3]);
```

database/migrations/2021_12_27_211938_drop_score_from_ratings_table.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('ratings', function (Blueprint $table) {
+            //
```
m-triassi, line 29: migrations should reverse themselves :) 
```suggestion
            $table->string('score')
```

database/migrations/2021_12_27_200344_add_categories_to_ratings_table.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('ratings', function (Blueprint $table) {
+            $table->dropColumn('length');
```
m-triassi, line 29:
```suggestion
            $table->dropColumn('categories');
```

PR #221, opened 2021-12-30
Install/Setup Laravel Backpack

### LINKED ISSUE #204: Install + Set-up Laravel-Backback
## Description

To give site admins access to reports for ratings a backend / CMS would be useful. This should allow for admins to manage and perform crud operations on most entities / models. Most / all models should be accessible via the CMS.

## Possible Implementation

Simply follow [the install guide](https://backpackforlaravel.com/docs/4.1/getting-started-basics) and get the package running, adding the basic CRUD actions for existing models

## Out of scope

N/A

Only Admins should be able to access this panel. There may be a way to restrict access via roles and policies. There may be more [details about authentication here](https://backpackforlaravel.com/docs/4.1/base-about#authentication-1)

RobertoNittolo:
## Description
Closes #204
Install Laravel Backpack
Add CRUD actions for existing models
Add a restriction to allow only site admins to be able to access Backpack

## How to access Backpack
Run the following command to install Backpack fully on your local `php artisan vendor:publish --provider="Backpack\CRUD\BackpackServiceProvider" --tag=public`
Create a user with a site admin role
Login in using a user account that has a site admin role
To access Laravel Backpack,  navigate to the /admin route.  https://localhost/admin
Once logged in you will see a sidebar with all the models accessible from Backpack.
Navigating to a model should allow for CRUD operations to be used


IanjhPhillips:
Is there any way to reduce the size of this PR?

RobertoNittolo:
> Is there any way to reduce the size of this PR?
The majority of the files were generated by installing Backpack. I think it will be inevitable that it will be a big pull request. I could perhaps make a new pull request for the roles/policies that still need to be implemented for this issue.

m-triassi:
Yeah it seems like this can be reduced in size by git ignoring that `public/packages` folder. It would mean though, that part of the installation / deploy procedure would be manually publishing those assets

https://backpackforlaravel.com/docs/4.1/faq#miscellaneous-1

Executive summary:
- add `/public/packages` to gitignore,
- run `git rm --cached public/packages`
- commit that 
- add a note somewhere that to install backpack fully on your local you need to run `php artisan vendor:publish --provider="Backpack\CRUD\BackpackServiceProvider" --tag=public`

m-triassi:
Just as a heads up, everything looks fine here but im still hesitating on the coverage; im not sure how its going to work with the exclusion rule. i'm going to give it a once over tomorrow, one last time.

m-triassi (approved):
Ok barring the coverage this seems fine -- i added an exclusion rule for the Admin "controllers" to code climate. What i'm going to do is merge, and see if the projects coverage drops 10%, if it does, i will revert the commit.

m-triassi:
Yep, Tanked the coverage, i think i can fix that though...

m-triassi:
😵 ok that was a bit messy, but it should be resolved now. Just for total transparency we are excluding coverage on the Admin controller name space because they don't actually represent functionality, but rather configuration. Thus just like we wouldn't write tests for the /config files, we also wont write tests for these.

PR #235, opened 2022-01-09
Revert "Install/Setup Laravel Backpack"

m-triassi:
Reverts DnD-Montreal/session-tome#221

prevent coverage drop before re-merging.


m-triassi:
Since this is a revert, we wont be doing a review

PR #239, opened 2022-01-10
Fix Item Entry Edit Bug

### LINKED ISSUE #209: Item Entry Edit Bug
## Description

When editing an entry which creates items, it is possible to delete and add items to the entry in such an order that the items to be deleted are not in fact deleted.

## Expected Behavior

Items to be deleted are deleted and items to be added are added.

## Actual Behavior

Items to be deleted are not deleted and items to be added are added.

## Steps to Reproduce

1. Open an entry edit form on an entry with items already attached to it
2. Choose to remove an item on the entry.
3. Create two new items on the entry.
4. Submit the edit form.

## Possible Fix

Remove the guard condition in item entry action which checks the count of items, and simply wipe all items from the entry each time the edit is passed and re create the items.

IanjhPhillips:
## Description
Closes #209 
Removes item count guard condition in CreateEntryItems action.
This guard condition made it possible to have items not be removed when they should be.
For example a user removes one item and then adds two new ones to a given entry as part of an edit, then with the guard condition the item to be removed would not have been removed.

### DIFF
--- a/app/Actions/CreateEntryItems.php
+++ b/app/Actions/CreateEntryItems.php
@@ -21,10 +21,7 @@ class CreateEntryItems
     {
         $out = collect();
 
-        if ($entry->items()->count() >= count($items)) {
-            // This is probably a "sync" so clear the existing items
-            $entry->items()->delete();
-        }
+        $entry->items()->delete();
 
         foreach ($items as $item) {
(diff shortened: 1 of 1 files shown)


m-triassi (approved):
lgtm 🤔

PR #257, opened 2022-01-11
Implement action to create dm rating upon entry store/update

### LINKED ISSUE #206: Attach ratings to GMs upon entry creation
## Description

Players should be able to create and attach ratings to entries and have that ratings be associated to the GM corresponding to the entry. 
Players should have the option to add a rating on either creation or editing of an entry.

## Possible Implementation

Use an action to implement creation of ratings and attaching them to an entry, similar to the Item Entry creation action

JasonGadoury:
## Description
Closes #206 

Adds an action that creates a DM rating when rating data is passed in EntryStoreRequest/ EntryUpdateRequest.

## How to test
Store:
1. Create an entry that includes a valid rating_data array within the EntryStoreRequest
2. Query for a Rating that matches the created entry's id
3. Check if decimal value of Rating's 'categories' field matches the binary value corresponding to the rating categories specified as true in request

Store:
1. Create an entry
2. Make an update to the entry that includes a valid rating_data array in the EntryUpdateRequest
2. Query for a Rating that matches the created entry's id
3. Check if decimal value of Rating's 'categories' field matches the binary value corresponding to the rating categories specified as true in request


app/Actions/CreateAndAttachRating.php
```diff
@@ -0,0 +1,40 @@
...
+            } elseif ($category == 'flexible' && $value) {
+                $rating->addCategory(FLEXIBLE_BITMASK);
+            } elseif ($category == 'friendly' && $value) {
+                $rating->addCategory(FRIENDLY_BITMASK);
+            } elseif ($category == 'helpful' && $value) {
+                $rating->addCategory(HELPFUL_BITMASK);
+            } elseif ($category == 'prepared' && $value) {
+                $rating->addCategory(PREPARED_BITMASK);
+            }
+        }
```
m-triassi, line 34: we can probably simplify this a ton by using a map (like [in this refactor](https://github.com/DnD-Montreal/session-tome/pull/220/commits/8c02a84d611540f772fb47878877c4cc9f601d14)) and string interpolation like this: 
```php
if ($value) {
    $key = strtoupper($category) . "_LABEL";
    $rating->addCategory($ratingMap[$key])
}
````

**Review by m-triassi — CHANGES_REQUESTED** [review]
here's a little suggestion for your action

File: app/Actions/CreateAndAttachRating.php (hunk 1 above)
Comment (m-triassi) [reply app/Actions/CreateAndAttachRating.php:34]: Also you're not referencing the class these constants come from... so you'd need to do: 
```suggestion
        foreach ($ratingData as $category => $value) {
            if ($category == 'creative' && $value) {
                $rating->addCategory(Rating::CREATIVE_BITMASK);
            } elseif ($category == 'flexible' && $value) {
                $rating->addCategory(Rating::FLEXIBLE_BITMASK);
            } elseif ($category == 'friendly' && $value) {
                $rating->addCategory(Rating::FRIENDLY_BITMASK);
            } elseif ($category == 'helpful' && $value) {
                $rating->addCategory(Rating::HELPFUL_BITMASK);
            } elseif ($category == 'prepared' && $value) {
                $rating->addCategory(Rating::PREPARED_BITMASK);
            }
        }
```

JasonGadoury: Both suggestions added.

app/Actions/CreateAndAttachRating.php
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Expect an array of rating_data, create Rating and modify categories as required
+     *
+     * @param Entry $entry
+     * @param array $ratingData
+     */
+    public function handle(Entry $entry, array $ratingData)
+    {
+        $rating = Rating::create([
```
m-triassi, line 22: to save the extra DB query we can just new an instance of it
```suggestion
        $rating = new Rating([
```

app/Http/Controllers/EntryController.php
```diff
@@ -53,26 +60,18 @@ public function store(EntryStoreRequest $request)
...
-        } elseif ($entryData->get('choice') == 'magic_item') {
-            // magic_item: attach item to character of choice, and set entry's levels to 0
-            $itemData = [$itemData[0] ?? []];
-            $entryData['levels'] = 0;
-        } elseif ($entryData->get('choice') == 'campaign_reward') {
-            // campaign reward: set levels = 0, no item(s), should contain custom note
-            $itemData = [];
-            $entryData['levels'] = 0;
-        }
+        //run tests to make sure this works
```
m-triassi, line 63: No need for this :)
```suggestion

```

app/Http/Controllers/EntryController.php
```diff
@@ -118,25 +122,18 @@ public function update(EntryUpdateRequest $request, Entry $entry)
...
```
m-triassi, line 125: same here
```suggestion

```

m-triassi (approved):
Nice work, good to see this one finally complete 👀

PR #275, opened 2022-01-12
Fix Integrity Constraint Error When Deleting Characters

### LINKED ISSUE #265: Character Delete Integrity Constraint
## Description

When deleting a character with entries and items attached, an error page is shown. 

## Expected Behavior

When clicking on the delete button the modal should appear, and then pressing delete should remove the character from you account, along with its entries and items

## Actual Behavior

An error is thrown and the character is not deleted.

error

## Steps to Reproduce

1. Login to an account with characters that have items and entries
2. Navigate to the Character's entry index
3. Click on the trashcan and confirm the delete.
4. Observe the error page

## Context

Characters can be created but not deleted :(

## Possible Fix

Add cascade constraints on relationships

m-triassi:
## Description
Closes #265 
Attempting to delete a character (or a user), raised an integrity constraint violation exception. This was because the character in question still had items, and ratings attached to their entries that were marked for deletion. 
This is solved by adding an observer that instead detaches entries from a character, leaving all the items and entries in tact. 

This approach prevents ratings or traded items from being deleted when a character is removed. 

Additionally soft deletes on users was added. The reasons are 2 fold: 
- Allows for users to be "banned"
- Allows a user to close their account without tossing away ratings and items they've created on the site.

Also adds a "default" account for ease of testing purposes. This account is automatically an admin. The credentials for this account are always `default@test.com` / `password`

## How to test
1. Migrate and seed your database
2. Login to an account with characters (See above example account)
3. Navigate to the character index
4. Pick a character and delete them
5. Observe that there is no error thrown
6. Open tinker (`php artisan tinker`)
7. Pick any user in the database (`$u = User::inRandomOrder()->first()`)
8. Observe their characters, entries, and items
9. Delete the user (`$u->delete()`)
10. Observe that their characters, entries, and items are still present in the database

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -5,6 +5,7 @@ namespace App\Models;
 use Illuminate\Contracts\Auth\MustVerifyEmail;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Foundation\Auth\User as Authenticatable;
 use Illuminate\Contracts\Auth\Authenticatable as AuthenticatableInterface;
 use Illuminate\Notifications\Notifiable;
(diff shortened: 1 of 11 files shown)


IanjhPhillips (approved):
Everything looks good here, ran the tests in my local and it behaved as expected.

The only thing I'd request is maybe adding a few more comments throughout the new migrations and the observer changes to explain what they do and why.

Otherwise I think its good to go!

PR #277, opened 2022-01-12
Data table search bar not filtering correctly

### LINKED ISSUE #270: Character Search Bar searches on incorrect fields
## Description

Searching for a character using the search bar on the character index should look at, ideally all fields, but it doesn't search all fields, making finding characters in a long list of characters difficult.

## Expected Behavior

Searching for a name in the search field should filter the results properly based on the character's name, class, race, etc...

## Actual Behavior

Typing in the search field only filters by a characters race.

## Steps to Reproduce

1. Sign in to an account with characters 
2. Goto the character index
3. Search for some content and notice that you can only search for words in the "Race" column

willyyhuang:
Closes #270 

This branch fixes filter in DataTable

### DIFF
--- a/resources/js/Components/Atom/DataTable.tsx
+++ b/resources/js/Components/Atom/DataTable.tsx
@@ -76,12 +76,13 @@ const DataTable = ({
     }, [data])
 
     // filter function
-    const filter = (row: DataType, target: string) => {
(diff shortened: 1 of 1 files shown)


m-triassi (approved):
This is small enough that i'll just merge this myself, it makes sense to me, good fix!

PR #279, opened 2022-01-13
#181 Character Actions Flow Improvements

### LINKED ISSUE #181: Character Management Flow Improvements
## Description
There are a few comments to be addressed about Character Management flow. Generally, they are pretty light requests. 

The below feedback is addressed 1 to 1 with the implementation section.

## Feedback
- Would be helpful if there was a dropdown feature for race/class/faction as to avoid typos for better organizing. 
- Will there be subclasses added? It would be great to have that eventually
- The scroll for the level/downtime isn't ideal since a slight movement and suddenly I'm level 20 with 4000000 downtime. I'd remove this and make it a manual entry for both.
- Sort would be helpful if the most recent characters added were first instead of on the second page
On the character page EDIT > UPDATE > would be ideal if it went right to SAVE instead of CONTINUE. Save a click

-Kate Ruscito

_Originally posted by @DndMtl in https://github.com/DnD-Montreal/session-tome/issues/159#issuecomment-973341191_

## Implementation Suggestions
- Back-end can add a list returned that has all races and classes, should be stored in either a config file, or a new model can be added / look-up table.
- Same as above subclasses can be addressed either way
- Pretty straight forwards, change to a basic input
- Back-end can return the characters in reverse order by default, which will resolve this. Additionally improved sorting via #164 will help alot here.

pascwhale:
Closes #181

## Description
- Change `Faction` input in the character creation form to a dropdown with acceptable values passed by backend.
- Remove arrows and scroll input from numerical fields `level` and `downtime` in the character creation form.
- Sort characters in `Characters` table by order of last updated first.


luigibesani (approved):
The faction drop down in the character create form works as intended.
The updated numerical fields look much less cluttered now that the arrows have been removed.
The sorting of the character table by last updated is a great way to go about organizing the order of a user's characters.

Good work pascal! 👍

resources/js/Components/Form/CharacterCreateForm.test.js
```diff
@@ -17,10 +17,24 @@ const editProps = {
         status: 'private',
     },
     editId: 0,
+    factions: [
```
willyyhuang, line 20: for coding standard, we should always make a constant when the code is repeated twice, in this case you should make a constant for factions but since this is a test file, i dont really care

resources/js/Components/Table/CharacterTable/CharacterTable.tsx
```diff
@@ -28,6 +29,7 @@ const CharacterTable = ({
     setEditId,
     setEditData,
 }: CharTablePropType) => {
+    data = orderBy(data, ['updated_at'], ['desc'])
```
willyyhuang, line 32: @m-triassi is it possible for BE to do the sorting on table rows initially?

m-triassi: Sure ill add that now!

m-triassi: @pascwhale i've added a sort query parameter to the character index, now if you pass `sort_by` in the query string you can supply one of the following fields to sort by it:  
"name,race,class,level,faction,downtime,updated_at"
you can also supply `sort_dir` as either "asc,desc", if `sort_dir` is not supplied it defaults to asc

so for example `/character?sort_by=name&sort_dir=desc` will sort by name descending. Or using ziggy's `route` function: `route('character.index', { "sort_by": "name", "sort_dir": "desc" })`

willyyhuang: this will be harder and should be worked on as a feature, for the scope of this task let's just do default sorting field for now

AnnieTran13 (approved):
- The dropdown for faction looks good and works as intended. Makes it easier to choose your faction. 
- The removal of arrow is good and works as intended. It would help a lot to apply that for other forms that require similar fields.
- The sorting by last updated is very good and works as intended. Good UX.

resources/js/Components/Table/CharacterTable/CharacterTable.tsx
m-triassi: I also added the default updated_at sorting with that feature.

pascwhale: Great, thank you for doing that and adding the matching tests! I removed frontend sorting @willyyhuang.

resources/js/Components/Form/CharacterCreateForm.test.js
pascwhale: Thanks, you're absolutely right! I've changed the test to use a single constant for factions.

willyyhuang (approved):

PR #281, opened 2022-01-14
Modify Campaign Model

### LINKED ISSUE #256: Modify Campaign Model
## Description

A campaign should have a unique code associated to it that is not its primary key. Players can use this code to join the campaign.

There should also be some helper methods to show if a given user is the GM or is a player.
(If the user is enrolled but does not have a character, they are the GM)

## Possible Implementation
Look into ways to create unique codes in laravel. You might be able to just add a column to the campaign migration called "code" or something which is a random integer. 
You could then make a campaign observer which creates a unique random string on a campaign when it is created. You would need to do something like this:
`Campaign::where('code', $code)->exists()` 
To figure out if the code you created is unique before associating it to the campaign.

For the campaign DM helper method, ~~you might consider making an [accessor](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor) on the campaign model which returns the user that is the DM~~. 
~~Otherwise you could also simply make a method is DM which returns true if the given user is the DM. (or both)~~
Add a [pivot value](https://laravel.com/docs/4.2/eloquent#working-with-pivot-tables) to the `campaign_users` table that flags a member of the campaign as a DM, boolean would work fine.

bcobo341:
## Description
Closes #256


app/Observers/CampaignObserver.php
```diff
@@ -0,0 +1,20 @@
...
+     * Handle the Campaign "created" event.
+     *
+     * @param  \App\Models\Campaign  $campaign
+     * @return void
+     */
+    public function created(Campaign $campaign)
+    {
+        $campaign->code = $campaign->generateCode();
+        $campaign->save();
+    }
```
m-triassi, line 19: technically you should be able to do 
```suggestion
    public function creating(Campaign $campaign)
    {
        $campaign->code = $campaign->generateCode();
    }
```

Saves you a database query; either way it works though

m-triassi (commented):
Just a small thing i noticed, while combing through the PRs, overall looking good though! when its ready for a full review, mark it as ready and re-request my review and ill take a look! :)

app/Observers/CampaignObserver.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    /**
+     * Handle the Campaign "created" event.
+     *
+     * @param  \App\Models\Campaign  $campaign
+     * @return void
+     */
+    public function creating(Campaign $campaign)
+    {
+        $campaign->code = $campaign->generateCode();
```
IanjhPhillips, line 17: We should probably check if another campaign exists that already has that code before assigning it. Could be a simple while loop until a unique code is generated. The probability of having a collision between two campaign codes is probably astronomically low, but it's still probably a good idea to be sure.

IanjhPhillips (changes requested):
Looking great overall so far!

app/Observers/CampaignObserver.php
m-triassi: Hah! we actually had a review offline in which we removed exactly that functionality; since we're using `microtime()` to generate the code then 2 users would have to create a campaign at the exact same microsecond to have a collision, i'd say that for uniqueness that's acceptable. I wanted to save on having to do more queries than necessary.

app/Models/Campaign.php
```diff
@@ -50,4 +50,9 @@ public function adventure()
     {
         return $this->belongsTo(\App\Models\Adventure::class);
     }
+
+    public function generateCode()
+    {
+        return dechex(microtime(true)*1000);
```
m-triassi, line 56:
```suggestion
        return dechex(microtime(true)*10000);
```
after i did some testing locally, microtime actually returns a number with 4 decimal places, so lets use all of them :)

m-triassi (approved):
I personally think this is just about clear @IanjhPhillips if you cant think of anything else you might need here for your PR then i'd say we can merge.

@bcobo341 please just address the merge conflicts?

app/Observers/CampaignObserver.php
IanjhPhillips: okay sounds good!

IanjhPhillips:
@m-triassi 
Added the new pivot value to the respective relationship functions on the Campaign and User models by adding 
`->withPivot('is_dm')`
to allow easier access to the pivot value.

IanjhPhillips (approved):
Tested everything in tinker, it all works as expected.

Great job @bcobo341 !

PR #283, opened 2022-01-14
Item form refactor

### LINKED ISSUE #282: Item form content refactor
## Description

Refactor item form content

willyyhuang:
Closes #282 

This branch changes `ItemForm` 
image

Clicking `Add Item` button will generate form fields that will set data correctly


luigibesani (approved):
I quite like the look of the refactored Item Form.
Adding and removing items is smooth like butter! 👍 
Good work @willyyhuang

pascwhale (approved):
The new form looks great and is much better in terms of UX! It's a lot better than having multiple action buttons at the bottom, so thank you for that. I took the liberty to change modify the case for the rarity option levels but everything else worked as expected. Nice job!

PR #284, opened 2022-01-15
Capitalize public/private status that appears on the character detail page

### LINKED ISSUE #267: Character exploration page shows the public status of a character in lowercase
## Description

Accessing a Character Detail / Exploration page of a character shows their public status in lowercase.

## Expected Behavior

`Public` or `Private` should be displayed instead of `public` or `private` under status.

## Actual Behavior

Only `public` or `private` are shown under status.

## Steps to Reproduce

1. Open any character detail page and notice the case of the status.

AnnieTran13:
Closes #267 

##  Description
The string public or private that appears on the character detail page now is now capitalized.

### DIFF
--- a/resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
+++ b/resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
@@ -19,6 +19,7 @@ const StyledTypography = styled(Typography)({
 
 const StyledStatus = styled(Typography)({
     color: '#a0a2a3',
+    textTransform: 'capitalize',
 })
 
 type CharDetailBoxPropType = {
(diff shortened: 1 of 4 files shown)


willyyhuang:
@m-triassi is this allowed? does BE allow status to be sent as `Private` or `Public`, if it is allowed i guess this is a solution

m-triassi:
> @m-triassi is this allowed? does BE allow status to be sent as `Private` or `Public`, if it is allowed i guess this is a solution

iirc any where queries we use in the back end are not case sensitive (The test suite would fail if they were), in an ideal world, the text _displayed_ would be upper case but the data returned to the back-end would be lowercase -- personally its not a big deal for me though.

willyyhuang:
@AnnieTran13 , let's send backend lowercase and have display text to be uppercase for the first letter

willyyhuang (changes requested):
refer to last comment

luigibesani (approved):
Privacy status is now capitalized. The use of a text-transform is convenient and clean. 
👍

m-triassi:
haha! HUGE PR!! 🚀 ill say this is fine for merge :)

PR #285, opened 2022-01-16
Implement Event Registration Route

### LINKED ISSUE #250: Create Event Registration Endpoint
## Description
Users need to be able to register for events. We need an endpoint that is given a session and
- check if the session has room
- registers the user if appropriate

m-triassi:
## Description
Closes #250 

Adds a store route that accepts a character and a session OR event ids. The route will register the selected character for either: a particular session in an event OR a random event in the session.

Adds a scope query that filters sessions by available seats. 

## How to test
Given this is just the backend implementation, without any way to interact via the front end, there isn't much you can do to test it. That said, you may want to test the newly added scope query.

1. Ensure your database is migrated and seeded
2. Get all Sessions and check that there are some with open seats and some without 
   - There is an accessor that you can use for this `$session->open_seats`
3. Call the scope query on sessions `Session::hasOpenSeats()->get()`. 
4. Observe that only Sessions with `open_seats` > 1 are returned. (`->pluck('open_seats', 'id')`)
5. If you'd like to filter by session with open seats from a particular event simply pass an event ID as the first parameter to the scope query `Session::hasOpenSeats(3)`. 

For testing of the end point in particular, feel free to add `dd()` statements and observe the output of the tests.


app/Http/Controllers/EventRegistrationController.php
```diff
@@ -0,0 +1,49 @@
...
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function store(RegistrationStoreRequest $request)
+    {
+        // this could probably be a request validator...
+        $data = $request->validated();
+
+        $character = Character::find($data['character_id']);
+        $request->user()->can('update', $character);
```
IanjhPhillips, line 27: Shouldn't this be in the request authorize method?

IanjhPhillips (approved):
Very well done, can't think of any edge cases that might've been missed.
Feel free to consider or ignore my change request

app/Http/Controllers/EventRegistrationController.php
m-triassi: Yeah you're right i just didn't want to be doubling up on the character fetch request, all good though shouldn't hurt performance in any tangible way, Fixed!

PR #290, opened 2022-01-16
Implement conditional button rendering for Character Detail page

### LINKED ISSUE #263: Hide update and export button on public profiles that you don't own
## Description

When visiting a public profile of a character that doesn't belong to the currently logged in user, the update and export buttons are visible and clickable which allows any other users to edit that character.

## Expected Behavior

The functionality of those buttons should be disabled and hidden when the public character profile is visited by other users to whom the character does not belong to.

## Actual Behavior

Other users can click on update and export buttons of a public character that don't belong to them. The drawer will show up but saving will not work.

## Steps to Reproduce

1. Go to a character profile page that is public and does not belong to logged in user
2. Click update or export button

luigibesani:
Closes #263 
## Previously
When visiting a public profile of a character that doesn't belong to the currently logged in user, the update and export buttons are visible and clickable which allows any other users to edit that character.

## Description
This PR makes use of a global user prop which is used to verify if the current existing user has to rights to access the Export and Update button of a certain character.
Only characters which belong to the existing users will display these two buttons on the CharacterDetail Page

## How to test this
1. Log into default@test.com
2. Access a character detail page of a character with an id >6
3. There will be no Export or Update buttons available as they are not rendered.

### DIFF
--- a/resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
+++ b/resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
@@ -3,9 +3,11 @@ import DownloadIcon from '@mui/icons-material/Download'
 import IosShareIcon from '@mui/icons-material/IosShare'
 import {Box, Button, Grid, Stack, Typography} from '@mui/material'
 import {Link} from 'Components'
-import React from 'react'
+import React, {useContext} from 'react'
 import styled from 'styled-components'
 import route from 'ziggy-js'
+
+import {UserContext} from '../../Layouts/ApplicationLayout/UserContext'
 
 const Img = styled('img')({
     margin: 'auto',
@@ -27,6 +29,11 @@ const StyledStatus = styled(Typography)({
 type CharDetailBoxPropType = {
     character: any
     setIsEditDrawerOpen: (payload: boolean) => void
+}
+
+function getUser() {
+    const user = useContext(UserContext)
+    return user
 }
 
 const CharacterDetailBox = ({character, setIsEditDrawerOpen}: CharDetailBoxPropType) => (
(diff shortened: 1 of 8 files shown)


resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
```diff
@@ -3,10 +3,12 @@ import DownloadIcon from '@mui/icons-material/Download'
 import IosShareIcon from '@mui/icons-material/IosShare'
 import {Box, Button, Grid, Stack, Typography} from '@mui/material'
 import {Link} from 'Components'
-import React from 'react'
+import React, {useContext} from 'react'
```
willyyhuang, line 6: as much as I like the use of a new hook, this is a decent approach but there are other better ways of doing this.
For example, we could use a state management library such as `redux`
Or since the backend is passing down `user` as a global object, you can just use a `Inertia` hook `usePage` to get the user object or even pass down from the parent component since it will exist there

resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
```diff
@@ -3,10 +3,12 @@ import DownloadIcon from '@mui/icons-material/Download'
...
 import styled from 'styled-components'
 import route from 'ziggy-js'
 
+import {UserContext} from '../../Layouts/ApplicationLayout/UserContext'
```
willyyhuang, line 10: don't like this import path

willyyhuang (changes requested):
Let's use inertia's `usePage`

resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
willyyhuang: https://inertiajs.com/shared-data

resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
luigibesani: On a second glance, I don't like it either @willyyhuang 😆

resources/js/Components/CharacterDetail/CharacterDetailBox.tsx
```diff
@@ -29,6 +31,12 @@ type CharDetailBoxPropType = {
     setIsEditDrawerOpen: (payload: boolean) => void
 }
 
+function getUser() {
```
willyyhuang, line 34: put this as an arrow function and within the component please. if you put it here the component will not re-render on state change

willyyhuang (changes requested):

willyyhuang (approved):

PR #233, opened 2022-01-09
Add Item Detail View page and Refactor ItemTable

luigibesani:
Closes #193
Closes #124
Closes #123 
## Description
- Update Character Detail Page
-- Add 'Item.index' route to CharacterDetail Page to render the Item Page through Inertia.

- Refactoring ItemTable Component
-- Implement the re-usable DataTable component  into the ItemTable Component in order to move away from a hard-coded table.
-- Add 'Item.show' route to ItemTable component to render ItemDetail Page when item name is clicked.

- Add Item Detail Component
-- Add CharacterItemCreateForm for the Drawer in the ItemDetail Component
-- Add ItemDetailBox Component to display Item data

- Edit Database Seeder to create Items associated to existing Characters
- Edit ItemController to prohibit viewing item's of a character which does not belong to the current User

PR #291, opened 2022-01-16
Item Tier bug

### LINKED ISSUE #274: Limit Tier for Items to 1-4
## Description

Tier field in Items should be between 1-4. 

## Expected Behavior

Tier field in Items should be limited to 1-4. 

## Actual Behavior

Currently, there is no limit to tier. The user can insert any number (no restrictions_

## Steps to Reproduce

1. Login to an account
2. Create an Entry (DM  or Character)
3. Once the first half is filled, add a magic item
4. In the "Tier" field manually type any number you'd like above 4
5. Observe in the database that the value chosen is not above 1 or below 4

## Possible Fix

There are 2 parts to fix this: add a validation rule on the `EntryStoreRequest` and `EntryUpdateRequest` that constrains the the tier to a value between 1-4, additionally add a mutator on on the model that will force the value to be below 4 and above 1.

bcobo341:
## Description
Adds a limit (integer from 1-4) to the tier field in Items 
Closes #274


m-triassi:
Just noticed this is in PR, @bcobo341 please make sure to add reviewers when a PR is ready for review aside from also moving out of draft!

m-triassi (approved):
this is pretty straight forwards, so looks all good to me. Also note that #274 was also partially fixed by some commits in #233

PR #292, opened 2022-01-19
Event Report Generation

### LINKED ISSUE #249: Event Report Generation
## Description
Admins need to be able to generate and download a report on a given event. This report should contain the event info as well as info about all characters who attended the event.

## Possible Implementation
Take a look at the GM rating report implementation for inspiration.

We might ebenefit from first getting a reference from the stakeholders of what they need in these reports.

IanjhPhillips:
## Description
Closes #249 
Creates a backpack endpoint that allows admin users to generate a report on a specific Event.
The report contains a list of all characters, their users, and the character's info.

## How to test
Make sure your database is migrated and seeded
Open tinker and run `$event = App\Models\Event::has('sessions')->first()`
Copy the title of that event
Log in as the admin user and head to /admin/event
Enter the title in the search bar
Click the "Generate Report" button
Your browser should attempt to download a csv. 
Check that the csv has character info and run the following command to compare the info
`$event->sessions()->with('characters')->get()`


app/Actions/GenerateEventReport.php
```diff
@@ -0,0 +1,66 @@
...
+        $columns = $this->prepareColumns();
+        $suffix = "event-report";
+
+        //get all characters that attended the event into one list
+        $characters = Character::whereHas(
+            'sessions',
+            function ($q) use ($event) {
+            $q->where('event_id', $event->id);
+        }
+        )->get();
```
m-triassi, line 35: I think the PSR standard for call backs like this is the following format, or atleast thats the one i've always followed

```suggestion
        $characters = Character::whereHas('sessions', function ($q) use ($event) {
            $q->where('event_id', $event->id);
        })->get();
```

app/Actions/GenerateEventReport.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Format the column text for the CSV to be
+     * @return string[]
+     */
+    private function prepareColumns(): array
+    {
+        $headings = ["User ID", "User Name", "Character ID", "Character Name", "Character Race", "Character Class", "Character Background"];
+        return $headings;
+    }
```
m-triassi, line 65: Since this is pretty static, this can probably just be a class const instead of a funciton; nbd either way though

m-triassi (commented):
Just some side comments, otherwise looks pretty solid so far!!

RobertoNittolo (commented):
Good work! Checked out the branch locally, was able to generate a report which contained the valid information!

app/Actions/GenerateEventReport.php
IanjhPhillips: weird, thats how I had it. Idk when it got changed but I definitely didnt format it that way.

app/Actions/GenerateEventReport.php
IanjhPhillips: yea youre probably right, I just had it this way cus I was anticipating that I might need to extend the functionality at some point

IanjhPhillips: after some consideration I decided to refactor it out to a const.

m-triassi (approved):
This is excellent! i'm going to approve, but i really think you should address the tier accessor

app/Models/Character.php
```diff
@@ -86,6 +86,52 @@ public function getGpAttribute()
...
+            case 17:
+            case 18:
+            case 19:
+            case 20:
+                $tier = 4;
+                break;
+            default:
+                $tier = 0;
+        }
+        return $tier;
```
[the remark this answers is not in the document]
IanjhPhillips: If this works ill change it to this

app/Actions/GenerateEventReport.php
```diff
@@ -0,0 +1,84 @@
...
+            foreach ($characters as $character) {
+                $attendanceCount += 1;
+                $user = $character->user;
+                if (!isset($data[$character->id])) {
+                    $data[$character->id] = [
+                        $attendanceCount,
+                        $event->id,
+                        $event->title,
+                        $session->adventure->id,
+                        $session->adventure->code,
```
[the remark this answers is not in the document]
IanjhPhillips: Yea fair ill make this change

m-triassi (approved):
Perfect! :))))

PR #293, opened 2022-01-19
Bulk entry delete

### LINKED ISSUE #264: Bulk Entry Delete
## Description

Allow multiple entries to be deleted by entries id

## Possible Implementation

Accept an array of ids in the same way the character delete function does on the `CharacterController`

JasonGadoury:
### Description
Closes #264  
Expands EntryController's destroy method to delete multiple entries passed as an array of entry IDs. 
Fixes bug in CharacterController's destroy method which allowed bulk character deletion by unauthorized user.

### How to Test
1. Create multiple entries with same user_id
2. Pass all entry ids as an array to destroy method
3. Check if specified entries have been successfully deleted from db

### DIFF
--- a/app/Http/Controllers/EntryController.php
+++ b/app/Http/Controllers/EntryController.php
@@ -154,9 +154,9 @@ class EntryController extends Controller
         if ($request->has('entries')) {
             $entries = Entry::whereIn('id', $data['entries'])->get();
 
-            foreach ($entries as $entry) {
-                if ($user->can('delete', $entry)) {
-                    $entry->delete();
+            foreach ($entries as $arrayEntry) {
+                if ($user->can('delete', $arrayEntry)) {
+                    $arrayEntry->delete();
                 }
             }
 
--- a/app/Http/Controllers/EntryController.php
+++ b/app/Http/Controllers/EntryController.php
@@ -155,9 +155,11 @@ class EntryController extends Controller
             $entries = Entry::whereIn('id', $data['entries'])->get();
 
             foreach ($entries as $entry) {
-                $user->can('delete', $entry);
-                $entry->delete();
+                if ($user->can('delete', $entry)) {
+                    $entry->delete();
+                }
             }
+
             return redirect()->route('entry.index');
         }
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/EntryController.php
```diff
@@ -144,9 +144,28 @@ public function update(EntryUpdateRequest $request, Entry $entry)
...
-        $entry->delete();
+        $user = $request->user();
+        $data = $request->validate([
+            'entries' => 'sometimes|array'
+        ]);
+
+        if ($request->has('entries')) {
+            $entries = Entry::whereIn('id', $data['entries'])->get();
+
+            foreach ($entries as $entry) {
```
IanjhPhillips, line 157: I would rename the **loop** `$entry` here to something slightly different just to avoid any confusion with the `$entry` **parameter** of the method. Maybe `$arrayEntry` just as an example.

tests/Feature/Http/Controllers/EntryControllerTest.php
```diff
@@ -674,15 +674,36 @@ public function update_campaign_reward_clears_items_and_levels()
      */
     public function destroy_deletes_and_redirects()
     {
-        $entry = Entry::factory()->create();
+        $entry = Entry::factory()->create(['user_id' => $this->user->id]);
+        $user = $entry->user;
```
IanjhPhillips, line 678: isn't this the same as just doing
`$this->user`? since thats what youre passing into the entry creation.

IanjhPhillips (approved):
Great work! I just had a few small comments, up to you if you wanna consider them.

m-triassi (approved):
I'd agree with ian here, up to you to address those things he's brought up. once you do, you can merge

tests/Feature/Http/Controllers/EntryControllerTest.php
JasonGadoury: Good point. I believe I added 677 later on and didn't bother to refactor. Will do.

PR #294, opened 2022-01-19
Campaign Filtering By Current User

### LINKED ISSUE #254: Campaign Index Filter
## Description

The Campaign controller index function should be filtered by a current user.
This way, the front end can show the current user only the campaigns they are a part of.

RobertoNittolo:
## Description
Closes #254 
Add filtering to the campaign controller index function to show campaign(s) that the logged-in user is a part of.

## How to test
1 Assign campaigns to a user
2 In CampaignController dd the following line in the index function `dd($campaigns = Auth::user()->campaigns()->get());`
3 Log into that user
4 Navigate to the campaign route `https://localhost/campaign`
5 Click through the provided collection, in the attributes section you should find the campaigns that were assigned to the user.

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -8,6 +8,7 @@ use App\Models\Campaign;
 use App\Models\User;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
+use Inertia\Inertia;
 
 class CampaignController extends Controller
 {
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -6,6 +6,7 @@ use App\Http\Requests\CampaignStoreRequest;
 use App\Http\Requests\CampaignUpdateRequest;
 use App\Models\Campaign;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
 
 class CampaignController extends Controller
 {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/CampaignController.php
```diff
@@ -5,19 +5,23 @@
...
     /**
      * @param \Illuminate\Http\Request $request
-     * @return \Illuminate\Http\Response
+     * @return \Inertia\Response
      */
     public function index(Request $request)
     {
-        $campaigns = Campaign::all();
+        $user = User::where('id', Auth::user()->id)->get();
+        $userCampaigns = $user->first()->campaigns()->get();
```
m-triassi, line 22: this can just be
```suggestion
        $campaigns = Auth::user()->campagins
```

app/Http/Controllers/CampaignController.php
```diff
@@ -5,19 +5,23 @@
...
 
-        return view('campaign.index', compact('campaigns'));
+        return Inertia::render('Campaign/Campaign', compact('userCampaigns'));
```
m-triassi, line 24: if the `Campaign/Campaign` page component doesn't exist, then you cant do this. You'd also need to change the test to expect an inertia response rather than a view.

Also if you take my suggestion from above, then you'd need to do

```suggestion
        return Inertia::render('Campaign/Campaign', compact('campaigns'));
```

m-triassi (commented):
Noticed some issues here

app/Http/Controllers/CampaignController.php
RobertoNittolo: I'll change it, good refactor opportunity!

app/Http/Controllers/CampaignController.php
RobertoNittolo: Alright, I'll change it back to return a view, since we don't have the campaign page component.

IanjhPhillips (approved):
Looks good!

PR #297, opened 2022-01-20
Sort Entries by Date Played

### LINKED ISSUE #296: Entry sorting is not in correct order
## Description

The sorting of entries in the character detail page is incorrect, as is the session counter. 

## Expected Behavior

Entries in the entry table should be sorted in reverse chronological order, from newest, to oldest. The "session" field should also count up from the oldest to the newest, when the adventures are the same. 

## Actual Behavior

Entries in the table are sorted by their primary keys. Session counter counts in the reverse order

## Steps to Reproduce

1. Login to an account with characters + entries where the same adventure is played atleast twice
2. navigate to the character in question's details page
3. Observe that the entries are in the incorrect order

## Possible Fix

Simply change some of the queries that serve data to the front end.

m-triassi:
## Description
Closes #296 

Adds proper sorting to lines in entry table. Lines are now sorted by `date_played` in reverse chronological order. Additionally counter for session now counts in the proper direction, labelling the oldest game as "1" and the newest game as "n" where n is the number of games played in 1 adventure.

## How to test
1. Login to an account with multiple entries for a character (preferably one with the same adventure across at least 2 entries)
2. Navigate to the Character's detail page
3. Observe that the entries are dated from newest to oldest.
4. Observe that the session attribute is also in descending order

### DIFF
--- a/app/Http/Controllers/CharacterController.php
+++ b/app/Http/Controllers/CharacterController.php
@@ -69,9 +69,13 @@ class CharacterController extends Controller
             abort(403);
         }
 
-        $entries = $character->entries()->with('adventure', 'items', 'rating')->get();
+        $entries = $character
+            ->entries()
+            ->with('adventure', 'items', 'rating')
+            ->orderBy('date_played', 'desc')
+            ->get();
         $factions = array_values(Character::FACTIONS);
-      
+
 
         return Inertia::render('Character/Detail/CharacterDetail', compact('character', 'entries', 'factions'));
     }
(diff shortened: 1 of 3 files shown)


RobertoNittolo (commented):
Pulled the branch locally, The dates were properly sorted from newest to oldest. Also looked at the session field, it properly counts up from oldest to newest when the adventures are the same. Good Job!

RobertoNittolo (approved):

PR #299, opened 2022-01-20
Rating Index

### LINKED ISSUE #216: Create Rating Index in Ratings Tab
## Description

The user should be able to see a list of GM ratings in the ratings tab. The user can filter event ratings or all ratings, sort by category rating, or by language.

## Possible Implementation

a more generic table (instead of data table) component could be implemented, that has Sortable columns.

## Additional Resources
[Mockups](https://www.figma.com/proto/5uF4bHcHFTm13iMyrBzMt0/Designs-and-Mockups?node-id=12%3A6&scaling=min-zoom&starting-point-node-id=942%3A33312&show-proto-sidebar=1)

pascwhale:
Closes #216 

## Description
Create a Rating table with a filter for League Event ratings. The filter uses a partial reload by calling the `Inertia.get` function with different parameters. What really makes it work is passing the `{preserveState: true, preserveScroll: true}` options that keep the component state and scroll position.


resources/js/Components/Table/RatingTable/RatingTable.tsx
```diff
@@ -0,0 +1,86 @@
+import DoneIcon from '@mui/icons-material/Done'
+import {Box, Chip, Typography} from '@mui/material'
+import {DataTable} from 'Components'
+import React from 'react'
+import {ReportedRatingData} from 'Types/reported-rating-data'
+
+type RatingTablePropType = {
+    ratingFilterStatus: boolean
```
willyyhuang, line 8: not a fan of the prop names here, let's go with something like `isFiltered`

resources/js/Components/Table/RatingTable/RatingTable.tsx
```diff
@@ -0,0 +1,86 @@
...
+    ratingFilterSetter: () => void
```
willyyhuang, line 9: same here, let's go with something like `setIsFiltered` since i supposed you are doing `useState`

resources/js/Components/Table/RatingTable/RatingTable.tsx
```diff
@@ -0,0 +1,86 @@
...
+    ratingFilterStatus,
+    ratingFilterSetter,
+    reportedRatings,
+}: RatingTablePropType) => {
+    const leftActions = [
+        <Chip
+            label='League Event Ratings Only'
+            onClick={() => ratingFilterSetter()}
+            deleteIcon={ratingFilterStatus ? <DoneIcon /> : undefined}
+            onDelete={ratingFilterStatus ? () => {} : undefined}
```
willyyhuang, line 23: `onDelete` here just does nothing?

resources/js/Components/Table/RatingTable/RatingTable.tsx
```diff
@@ -0,0 +1,86 @@
...
+            property: null,
+            title: 'Prepared',
+            render: (row: ReportedRatingData) => (
+                <Typography>{row?.total_ratings?.PREPARED ?? '0'}</Typography>
+            ),
+        },
+    ]
+
+    return (
+        <Box>
```
willyyhuang, line 72: is `Box` needed here?

resources/js/Pages/Rating/Rating.tsx
```diff
@@ -0,0 +1,42 @@
...
+type RatingPropType = {
+    users: ReportedRatingData[]
+}
+
+const Rating = ({users}: RatingPropType) => {
+    const [isRatingFilteredForEvent, setIsRatingFilteredForEvent] =
+        useState<boolean>(false)
+    const filterTable = () => {
+        const newFilterState = !isRatingFilteredForEvent
+        setIsRatingFilteredForEvent(newFilterState)
```
willyyhuang, line 21: don't really understand the 2 lines here

willyyhuang (changes requested):

resources/js/Components/Table/RatingTable/RatingTable.tsx
pascwhale: I tested this out. I believe that MUI does not show the icon next to the chip if there is no `onDelete` prop being passed. It's weird but it works.

resources/js/Components/Table/RatingTable/RatingTable.tsx
pascwhale: You're right, I'll remove it.

resources/js/Pages/Rating/Rating.tsx
pascwhale: For some reason the state hook is not being called before moving on to the next command which needs the boolean in its right state..

willyyhuang (approved):

m-triassi:
Looks like this is ready for merge!! :)

PR #305, opened 2022-01-23
Create Campaign Join Endpoint

### LINKED ISSUE #255: Create Campaign Join Endpoint
## Description

Users should be able to join a campaign if given a campaign join code

bcobo341:
## Description
Closes #255
Player and GM can join a campaign


app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
+    {
+        $data = $request->validate([
+            'character_id' => "required|exists:characters,id",
```
IanjhPhillips, line 22: Having character ID be required here implies that a user cannot join a campaign as the GM, and so the creator of the campaign must be the GM. Is this a restriction we want to impose?

There are pros and cons for it (such as ensuring there's only ever one GM).
However, I think we should allow for users to join as the GM, if for whatever reason a player is the creator of the campaign.

That would mean character_id here should be 'sometimes'

IanjhPhillips: you would also then be expecting a third request parameter being 'is_dm' or something along that line.

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,42 @@
...
+            'code' => "required|string"
+        ]);
+
+        $user = User::find(Auth::id())->first();
+        $character = Character::findOrFail($data['character_id'])->first();
+        $campaign = Campaign::where('code', '=', $data['code']);
+
+        if ($campaign->exists()) {
+            $campaign = $campaign->first();
+            $user->campaigns()->attach($campaign);
```
IanjhPhillips, line 32: regardless of your decision on my above comments of the request validation rules, I think you should be explicitly passing in the pivot value here as false if the user you're attaching is a player.
`$user->campaigns()->attach($campaign, ['is_dm' => false])`

IanjhPhillips (changes requested):
Looks good so far!
I think it would be good to allow users to join as either a player or a GM, meaning we'd need to expect an 'is_gm' parameter in the request.
We'd also only be attaching a character if is_gm is false, and also include that value in the pivot when attaching the user to the campaign.

app/Http/Controllers/CampaignRegistrationController.php
bcobo341: Right, I wasn't sure if a user could join a campaign as a GM/dm. I can make the adjustment if that is the idea.

app/Http/Controllers/CampaignRegistrationController.php
bcobo341: is_dm automatically defaults to false, however, I can explicitly specify it.

IanjhPhillips: ah okay i didnt know this, in that case its fine

app/Http/Controllers/CampaignRegistrationController.php
IanjhPhillips: Yea i think we should leave the option for users to join as GM

IanjhPhillips:
Also remember to add new endpoints to the wiki

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,42 @@
...
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+use App\Models\Campaign;
+use App\Models\User;
+use App\Models\Character;
+use Illuminate\Support\Facades\Auth;
+
+class CampaignRegistrationController extends Controller
+{
```
m-triassi, line 12: we're going to want to add a `create(Request $request) {...}` function + route here, so we can show a form that will allow a user to choose what character will be joining the campaign or if they'll be joining as a GM. The Create route should receive all the logged in user's characters. Also you should pass along the `code` as prop to the view that you create to help the front-ends

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,42 @@
...
+    public function store(Request $request)
+    {
+        $data = $request->validate([
+            'character_id' => "required|exists:characters,id",
+            'code' => "required|string"
+        ]);
+
+        $user = User::find(Auth::id())->first();
+        $character = Character::findOrFail($data['character_id'])->first();
+        $campaign = Campaign::where('code', '=', $data['code']);
```
m-triassi, line 28: slightly irrelevant, but the operator parameter isn't really needed, so you can just do
```suggestion
        $campaign = Campaign::where('code', $data['code']);
```

m-triassi (commented):
I mostly agree with ian here, added some small comments as well

app/Http/Controllers/CampaignController.php
```diff
@@ -37,7 +39,20 @@ public function create(Request $request)
      */
     public function store(CampaignStoreRequest $request)
     {
-        $campaign = Campaign::create($request->validated());
+        $data = $request->validated();
+
+        $campaign = Campaign::create($data);
+
+        //user joins capaign
+        $user = User::find(Auth::id())->first();
```
m-triassi, line 47: Theres a short hand that does this for you
```suggestion
        //user joins capaign
        $user = Auth::user()
```

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public function create(Request $request)
+    {
+        $characters = Auth::user()->characters;
+
+        $code = $request->validate([
+            'code' => "sometimes|string"
+        ])['code']?? null;
+
+        $campaign = Campaign::where('code', $code)->first();
```
m-triassi, line 26: lets also eager load the users and characters that are already part of the campaign, so the front end will be able to show the characters that have already joined, also we can make the page 404 if they didn't enter the right code with `->firstOrFail()`
```suggestion
        $campaign = Campaign::with('characters', 'users')->where('code', $code)->firstOrFail();
```

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $code = is_null($campaign) ? $campaign : $campaign->code;
```
m-triassi, line 28: Im not sure what this line is doing, in all honesty, we can probably just remove it
```suggestion
```

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,66 @@
...
+     * @param \Illuminate\Http\Request $request
+     * @return \Inertia\Response
+     */
+    public function create(Request $request)
+    {
+        $characters = Auth::user()->characters;
+
+        $code = $request->validate([
+            'code' => "sometimes|string"
+        ])['code']?? null;
```
m-triassi, line 24: PSR-1 coding standards

```suggestion
        ])['code'] ?? null;
```

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
+    {
+        $data = $request->validate([
+            'character_id' => "sometimes|exists:characters,id",
```
m-triassi, line 42: knowing the front ends, its likely they're going to pass `"character_id": null`, so might as well just add 
```suggestion
            'character_id' => "nullable|sometimes|exists:characters,id",
```

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,66 @@
...
+            'code' => "required|string"
+        ]);
+
+        $user = User::find(Auth::id())->first();
```
m-triassi, line 46: you can just use the short hand here again 
```suggestion
        $user = Auth::user();
```

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,66 @@
...
+                $character = Character::findOrFail($data['character_id'])->first();
+                $character->campaigns()->attach($campaign);
+            } else {
+                $user->campaigns()->attach($campaign, ['is_dm' => true]);
+            }
+        } else {
+            return back()->withErrors([
+                'code' => "The provided code does not exist!"
+            ]);
+        }
```
m-triassi, line 62: id probably leverage first or fail here for less control flow, otherwise, its a pretty good implementation
```suggestion
        $campaign = Campaign::where('code', $data['code'])->firstOrFail();

        if ($request->has('character_id')) {
            $user->campaigns()->attach($campaign, ['is_dm' => false]);
            $character = Character::findOrFail($data['character_id']);
            $character->campaigns()->attach($campaign);
        } else {
            $user->campaigns()->attach($campaign, ['is_dm' => true]);
        }
```

m-triassi (changes requested):
Really close! you can just use the suggestions, but ill leave fixing the tests to match up to you :)

tests/Feature/Http/Controllers/CampaignControllerTest.php
```diff
@@ -76,9 +77,37 @@ public function store_saves_and_redirects()
...
+        $response->assertRedirect(route('campaign.index'));
+        $response->assertSessionHas('campaign.id', $campaign->id);
+        $this->assertDatabaseCount('campaign_user', 1);
+        $this->assertDatabaseHas('campaign_user', ['user_id' =>$this->user->id, 'campaign_id' =>$campaign->id, 'is_dm' =>true]);
+    }
+
+    /**
+     * @test
+     */
+    public function store_saves_character_and_redirects()
```
IanjhPhillips, line 101: shouldnt this be `store_saves_campaign_...` not `store_saves_character_...` ?

IanjhPhillips (changes requested):
Nothing left I can see to change functionality wise, though I would consider changing the name of the store test case in CampaignTest.

tests/Feature/Http/Controllers/CampaignControllerTest.php
m-triassi:
```suggestion
    public function store_saves_campaign_and_redirects()
```

IanjhPhillips (approved):

app/Http/Controllers/CampaignController.php
```diff
@@ -37,7 +39,20 @@ public function create(Request $request)
...
+        $data = $request->validated();
+
+        $campaign = Campaign::create($data);
+
+        //user joins capaign
+        $user = Auth::user();
+
+        if ($request->has('character_id')) {
+            $user->campaigns()->attach($campaign, ['is_dm' => false]);
+            $character = Character::findOrFail($data['character_id'])->first();
```
m-triassi, line 51: You can omit the `first()` here
```suggestion
            $character = Character::findOrFail($data['character_id']);
```

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $characters = Auth::user()->characters;
+
+        $code = $request->validate([
+            'code' => "sometimes|string"
+        ])['code'] ?? null;
+
+        $campaign = Campaign::with('characters', 'users')->where('code', $code)->first();
+
+        return Inertia::render('Campaign/Detail/CampaignDetail', compact('characters', 'campaign'));
```
m-triassi, line 27: it'd also pass down code here, will be convenient for the front-ends later
```suggestion
        return Inertia::render('Campaign/Detail/CampaignDetail', compact('characters', 'campaign', 'code'));
```

m-triassi (approved):
Nice work! This is ready to go; ill just commit those 2 small things for ya first

PR #323, opened 2022-01-25
Adventure field type ahead

### LINKED ISSUE #219: Add type-ahead to Adventure field on Entry
## Description

Searching for an adventure to attach to an entry should be quick. We are seeing some lag when searching the large list of adventures on the front end.

When searching for an adventure to attach to an entry, a request to a new endpoint should be made that passes the search term. The endpoint should return a limited result set based on the search term. 
 
## Expected Behaviour

- Begin typing an adventure title in the entry form
- Request is sent to the backend
   - Debounce if a user is typing quickly
- Backend returns list based on search term

## Actual Behaviour

- Entire list of adventures is returned
- Front end is filtering list based on input

## Steps to Reproduce

1. Login
2. Create an entry (and a character if one is not present)
3. Begin typing an adventure title
4. Observe the slow down

## Possible Fix

Implement new Adventure search endpoint (fuzzy search on title and code), add debounce search select that sends request on every keypress.

willyyhuang:
Closes #219 

Add searchable autocomplete to adventure fields


pascwhale (approved):
Everything works beautifully @willyyhuang!

willyyhuang:
with the change of `adventure_id` to `adventure.id` what would the server side validation `errors.adventure_id` change to?
`errors.adventure.id` ? `errors.adventure` @m-triassi

m-triassi:
> with the change of `adventure_id` to `adventure.id` what would the server side validation `errors.adventure_id` change to? `errors.adventure.id` ? `errors.adventure` @m-triassi

I imagine it would be `errors.adventure.id`, yes, but i cant quite say for sure

IanjhPhillips (approved):
Everything looks great!
I tested locally and was able to filter the adventure field with the type ahead, it worked perfectly. I wasn't able to do the same for the GM field though. I dont know if that was supposed to work yet or not.

Also, the filterable trait looks good, and I can imagine it'll definitely be useful in the future. It also is definitely an implementation of a design pattern which we could include in our documentation.

willyyhuang:
@IanjhPhillips i can add game masters in this branch, it just requires the backend to pass down game master correctly in character detail if it is not there already

m-triassi:
@IanjhPhillips @willyyhuang lets save that for next sprint, since i feel the implementation will be non trivial and is technically outside the scope of this issue

PR #325, opened 2022-01-27
Disable scroll for numerical inputs in forms

### LINKED ISSUE #304: Numerical inputs in forms should disable scroll
## Description

Accidentally scrolling within a default numerical field in MUI might change its value without the user noticing. The arrows on the right side of the field also take space. This includes the `Length` /`Levels`/ `GP` fields in Entry and DM Entry forms, as well as `Tier` in the Item form.

## Expected Behavior

Use the changes made in the #181 Character Flow Improvement issue, which disables scrolling as well as the incrementing arrows within the field.

## Actual Behavior

Scrolling changes the value and arrows take space within the field.

## Steps to Reproduce

1. Open the Create Entry or DM Entry form.
2. Scroll over the Length/Levels/GP fields to see the values changing.
3. Notice the space the arrows take, especially in the edit version of those forms.

## Possible Fix

Here is an example of a fix.
The prop `type='number'` is removed from the `Textfield` component.
Props `inputMode: 'numeric',` and `pattern: '[0-9]*',` are added to the `InputProps` props.
A condition is checked in the `onChange` prop to set the value to `0` when the field value is erased. Additional checks can also be done depending on the field logic.

# Other

AnnieTran13:
Closes #304 

## Description

- Disabled scroll (removed up/down arrow buttons in text field) for length, levels and gp inputs in entry form (EntryCreateForm) and in DM entry form (DMEntryCreateForm).

- Changed the tier textfield to a dropdown using MUI textfield select property.

- Updated test that uses the ItemForm and ItemEditForm

### Additional note
- Length field: Can't be below 0.
- Levels field: Can't be below 0, can't be above 20.
- GP field: Can be below 0. (When the character loses money during a play)


resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -64,6 +64,13 @@ const StyledGrid = styled(Grid)`
     margin-bottom: 16px;
 `
 
+const StyledGPTextField = styled(TextField)({
+    '& input::-webkit-clear-button, & input::-webkit-outer-spin-button, & input::-webkit-inner-spin-button':
+        {
+            display: 'none',
+        },
+})
```
pascwhale, line 72: Hey @AnnieTran13, what is this doing exactly?

resources/js/Components/Form/DmEntryCreateForm.tsx
```diff
@@ -165,33 +185,49 @@ const DmEntryCreateForm = ({
...
-                        inputProps: {
-                            min: 0,
-                        },
-                    }}
                     value={data.gp.toString()}
-                    onChange={(e) =>
-                        setData('gp', parseFloat(parseFloat(e.target.value).toFixed(2)))
-                    }
+                    onChange={(e) => {
+                        const value = parseFloat(parseFloat(e.target.value).toFixed(2))
```
pascwhale, line 224: Can you explain why you have nested calls for the `parseFloat` function?

pascwhale (changes requested):
Did you forget to modify the field for the Item create form's tier field? I believe that was also in the scope of #304.

resources/js/Components/Form/DmEntryCreateForm.tsx
willyyhuang: same question

resources/js/Components/Form/EntryCreateForm.tsx
```diff
@@ -150,14 +157,27 @@ const EntryCreateForm = ({
...
                     InputProps={{
                         inputProps: {
                             min: 0,
+                            inputMode: 'numeric',
+                            pattern: '[0-9]*',
                         },
                     }}
                     value={data.length.toString()}
-                    onChange={(e) => setData('length', parseInt(e.target.value))}
+                    onChange={(e) => {
```
willyyhuang, line 171: if multiple fields have the same or almost identical `onChange` function, it is a better idea to extract them and make a re-usable component
in this case, we should probably make a re-usable component for float `textField`

resources/js/Components/Form/DmEntryCreateForm.tsx
AnnieTran13: So I looked up how to disable the arrow buttons on a textfield of type number. This is basically just hiding the arrow buttons. The problem with the GP textfield is that we can't remove the type like you did with the other textfields. I tried a lot of different things to keep GP as a float without the type set to number and it just doesn't work.

resources/js/Components/Form/DmEntryCreateForm.tsx
AnnieTran13: I think @luigibesani implemented that so I don't know.

pascwhale (approved):
Everything looks good! The dropdown added for the Tier field in Item forms works too.

willyyhuang (changes requested):

willyyhuang (changes requested):
some small changes needed

PR #327, opened 2022-01-27
Fix entry fields saving / displaying improperly

### LINKED ISSUE #326: Length, levels and gp in the edit drawer of dm entries and entries are not showing the proper value
## Description

The length, levels and gp field in the edit drawer of the edit drawer of DM entries and entries are not showing the right value. 

## Expected Behavior

After creating/editing length, levels or gp of an entry, the data should be updated on the edit drawer if you were to edit again.

## Actual Behavior

After creating/editing length, levels or gp of an entry, the data is NOT updated on the edit drawer if you were to edit again.

## Steps to Reproduce

1. Go to a character detail page.
2. Click on the the edit icon to open the drawer
3. length, levels and gp won't show the right value (unless it's 0)

## Possible Fix

Levels and gp are a frontend problem and length is a backend problem.

# Other

m-triassi:
## Description
Closes #326 

Formerly when editing or saving an entry (DM or character) the fields Length, GP, and Levels would always display 0. Additionally length would not be saved in the database.

This Fixes those issues by properly targeting the passed prop data, and adding missing validation rules for Length

## How to Test
1. Login to a user with Characters / Entries
2. Edit the Length / Levels / GP
3. Make sure the values shown are correct


pascwhale (commented):
It worked for character Entries but not for DM Entries.

willyyhuang (approved):

IanjhPhillips:
I tested it locally and ran into some issues

The edit drawers for both character entries and GM entries still display level and gp fields as 0.

Also, I tried creating an entry to make sure it didnt have zero in those fields, and I was given a form validation error on the adventure field. It said adventure id is required. I don't know why I'm getting this error as I picked an adventure from the drop down.

pascwhale (approved):
Editing now works for both character and DM entries. Great work everyone!

m-triassi:
@IanjhPhillips it sounds like you didn't do a `yarn dev` before testing?

PR #345, opened 2022-01-29
Implement performance tests

RobertoNittolo:
## Description 
The former Method for the performance tests was not working properly. It was not allowing the locust script to properly create characters. We figured out that we would need to use API tokens for it to work properly.

-Add a Locust Controller to be able to generate an API token 
-Add API routes for character create and delete
-Add API route for Token generation
-Add Tests for the Locust Controller


m-triassi (approved):
Ok this looks alright, we'll need to remember to delete this extra code after release 3, would you mind open an issue for that?

PR #348, opened 2022-01-30
Scope rating report to league

JasonGadoury:
### Description
Adds button to league admin panel to generate a rating report for a specific league. 
Modifies current GenerateRatingReport action to handle a passed league_id in order to scope ratings to that league.

### To Test

1. Create multiple ratings from events belonging to at least 2 different leagues
2. Run the GenerateRatingReport action passing parameters (true, $league) where $league is the league created that you would like to export the ratings for (pass the object not the league id)
3. Check if report contains only ratings from specified league


JasonGadoury:
Need to pass the league_id to the action. Will fix

m-triassi (approved):
This works alright, clear to merge i'd say

JasonGadoury:
@m-triassi thanks!

PR #356, opened 2022-01-31
Hotfix on auto complete for adventure field in bulk entry

willyyhuang:


m-triassi (approved):
I'm going to make an executive decision and just merge this, it seems to function fine

PR #388, opened 2022-02-08
Fix Bulk Entry Date skipping bug

### LINKED ISSUE #332: Bulk Entry creation skips a week
## Description

Bulk Entry Creation Occasionally Skips a week during creation, if a range of 5 weeks is chosen, there may only be 4 entries created.

## Expected Behavior

When creating Character entries in bulk, if the chosen date range spans 5 weeks, and the frequency is set to once / week, there should be 5 total entries created. 

## Actual Behavior

When using the bulk entries utility, that spans 5 weeks, it will intermittently skip a week. 
unknown

## Steps to Reproduce

It's currently unclear what causes this bug. 
1. Login to an account with a character
2. Open the bulk entry utility 
3. Choose some multi-week period and a frequency of once per week
4. Observe if the correct amount of entries are created

## Possible Fix

This is likely an issue in the Bulk Entry controller and its date logic

m-triassi:
## Description
Closes #332 

Refactors Bulk entry Store to use carbon's `CarbonPeriod` class rather than manually calculating the date ranges. This both fixes the bug where the 3rd week in a span of 4 weeks was being skipped, as well as makes the function more readable.

## How to test
1. Login to the default account
2. (Optional) delete all the entries on a chosen character
3. Open the bulk entry form
4. Choose a date range spanning multiple weeks, set the frequency to "once per week"
5. Hit create
6. Observe that the correct number of entries for the selected number of weeks was created
7. Experiment with other date ranges and frequencies.

### DIFF
--- a/app/Http/Controllers/BulkEntryController.php
+++ b/app/Http/Controllers/BulkEntryController.php
@@ -6,6 +6,7 @@ use App\Http\Requests\BulkEntryStoreRequest;
 use App\Models\Adventure;
 use App\Models\Character;
 use Carbon\Carbon;
+use Carbon\CarbonPeriod;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
 use Inertia\Inertia;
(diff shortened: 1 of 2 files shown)


IanjhPhillips (approved):
Tested it locally and works as expected.

The code looks a lot cleaner now too with Carbon

app/Http/Controllers/BulkEntryController.php
```diff
@@ -39,23 +40,24 @@ public function create(Request $request)
...
-            // N+1 consider refactor?
-            $entry->save();
-        }
+        $character->entries()
+            ->where('created_at', ">=", now())
+            ->get()
+            ->each(function ($entry, $key) use ($dates) {
+                $entry->date_played = $dates[$key];
+                $entry->save();
+            });
```
JasonGadoury, line 60: Neat refactor of the original foreach loop's functionality within the query.

JasonGadoury (approved):
This fix looks pretty good.

PR #390, opened 2022-02-08
Trade Detail Response

### LINKED ISSUE #377: Modify Trade Detail Response
## Description

If the current user did not create the trade, the detail view should include info about the character and user posting the trade. This allows front end to include a link to the character or user as well as their name.

If the current user did create the trade, the detail view should include info about any offers they have received on their trade.

## Possible Implementation

Add data to the compact portion of the response.

IanjhPhillips:
## Description
Closes #377 
Modifies the TradeController's show endpoint to return more data. Also, data returned depends on if the current user is the creator of the given trade or not.

If they are not, they are only shown the character, user, and trade item info. This allows the user to view the character and item that pertain to the trade.

If they are, they are shown the above info as well as offers that have been made against their trade. These offers come in the for of items with their associated character loaded as well, so that the user can audit the character.

### DIFF
--- a/app/Http/Controllers/TradeController.php
+++ b/app/Http/Controllers/TradeController.php
@@ -6,6 +6,7 @@ use App\Http\Requests\TradeStoreRequest;
 use App\Http\Requests\TradeUpdateRequest;
 use App\Models\Trade;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
 
 class TradeController extends Controller
 {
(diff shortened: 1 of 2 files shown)


willyyhuang:
There is a code smell according to sonarcloud

app/Http/Controllers/TradeController.php
```diff
@@ -49,7 +50,15 @@ public function store(TradeStoreRequest $request)
      */
     public function show(Request $request, Trade $trade)
     {
-        return view('trade.show', compact('trade'));
+        $currentUserID = Auth::user()->id;
```
m-triassi, line 53: Little nit-picky but you can just do 
```suggestion
        $currentUserID = Auth::id();
```
here :)

m-triassi (approved):
pretty straight forward! This seems to work alright :) ill let you decided if you want to commit that suggestion i made

PR #391, opened 2022-02-09
Rating step skipped when the gm field is empty

### LINKED ISSUE #308: Entry form should skip rating step if the GM field is empty
## Description

The entry form should not allow the user to go to the Rating step of the form if no GM was entered in the first step.

## Expected Behavior

The entry form should automatically skip the Rating step when moving on to the next step. The form should also go back to the first page when clicking on the `Previous` button on the third page if no GM is entered.

## Actual Behavior

The entry form is shown when hitting next despite the GM field being empty. 

## Steps to Reproduce

1. Open the Character Detail page.
2. Create an Entry.
3. Click next without entering a GM.

## Possible Fix

# Other

AnnieTran13:
Closes #308 

If GM field is empty
- When clicking continue on the first step of Entry form, skips to 3rd step.
- From the 3rd step, when clicking previous, goes back to first step.

### DIFF
--- a/resources/js/Components/Form/EntryCreateForm.tsx
+++ b/resources/js/Components/Form/EntryCreateForm.tsx
@@ -306,23 +306,15 @@ const EntryCreateForm = ({
     const stepThreeFooter = (
         <StyledGrid container spacing={4}>
             <Grid item xs={4}>
-                {data.dungeon_master === '' ? (
-                    <Button
-                        fullWidth
-                        onClick={() =>
-                            type === 'Create' ? setActiveStep(0) : setActiveStep(0)
-                        }>
-                        Previous
-                    </Button>
-                ) : (
-                    <Button
-                        fullWidth
-                        onClick={() =>
-                            type === 'Create' ? setActiveStep(1) : setActiveStep(0)
-                        }>
-                        Previous
-                    </Button>
-                )}
+                <Button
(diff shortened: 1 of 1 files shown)


resources/js/Components/Form/EntryCreateForm.tsx
```diff
@@ -304,7 +311,9 @@ const EntryCreateForm = ({
                 <Button
                     fullWidth
                     onClick={() =>
-                        type === 'Create' ? setActiveStep(1) : setActiveStep(0)
+                        type === 'Edit' || data.dungeon_master === ''
```
willyyhuang, line 314: you can just use `!data.dungeon_master` if you are looking for empty string, undefined or null

willyyhuang:
@m-triassi just to confirm, the GM field is not required anymore right?

m-triassi:
> @m-triassi just to confirm, the GM field is not required anymore right?

@willyyhuang it shouldn't be no, if it is left empty, then the `dungeon_master` and `dungeon_master_id` will be set to null

willyyhuang (approved):

resources/js/Components/Form/EntryCreateForm.tsx
AnnieTran13: Done!

PR #392, opened 2022-02-10
Gamemaster field update

### LINKED ISSUE #295: Dungeon/game master field should be an autocomplete field.
## Description

When filling out an entry form, a user should be able to search the system for the DM that ran their game. 

## Expected Behavior

Typing the name of a DM should reveal a drop down of selections that are close the the query string, and allow the user to pick from that list. if a user picks a DM from this list then the user chosen should have their **id** passed in the form as `dungeon_master_id`, if no user is selected from the drop down, then they typed string should be passed as `dungeon_master`.

## Actual Behavior

The field currently only accepts the string representation of the dm's name and passes it as `dungeon_master`. The drop down select doesn't render, nor can a user have their ID passed as `dungeon_master_id`.

## Steps to Reproduce

1. Sign in to an account
2. Create a character
3. Create an entry for that character
4. Observe in the form that typing the name of a different user in the "game master" field has no effect.

## Additional Notes

It's been mentioned that we should the dungeon/game master field should be an autocomplete field. We will need to create another issue for that.

_Originally posted by @AnnieTran13 in https://github.com/DnD-Montreal/session-tome/pull/237#discussion_r787315392_

willyyhuang:
Closes #295 

Changes game master field to use autocomplete component, however, it should be allowed to input purely string

This affects: 
Character entry creation, edit

### DIFF
--- a/app/Http/Controllers/CharacterController.php
+++ b/app/Http/Controllers/CharacterController.php
@@ -73,7 +73,8 @@ class CharacterController extends Controller
         $search = $request->get('search', "");
 
         $entries = $character->entries()
-            ->with('adventure', 'items', 'rating', 'dungeonMaster')
+            ->with('adventure', 'items', 'rating')
+
             ->orderBy('date_played', 'desc')
             ->get();
         $factions = array_values(Character::FACTIONS);
--- a/app/Http/Controllers/CharacterController.php
+++ b/app/Http/Controllers/CharacterController.php
@@ -73,7 +73,7 @@ class CharacterController extends Controller
         $search = $request->get('search', "");
 
         $entries = $character->entries()
-            ->with('adventure', 'items', 'rating')
+            ->with('adventure', 'items', 'rating', 'dungeonMaster')
             ->orderBy('date_played', 'desc')
             ->get();
         $factions = array_values(Character::FACTIONS);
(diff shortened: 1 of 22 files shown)


app/Http/Controllers/CharacterController.php
```diff
@@ -74,6 +74,7 @@ public function show(Request $request, Character $character)
 
         $entries = $character->entries()
             ->with('adventure', 'items', 'rating')
+
```
IanjhPhillips, line 77: why

IanjhPhillips (commented):
All the backend and cypress changes make sense, I cant see any glaring issues.

The only thing im unsure of is in the entry controller edit function, shouldnt we be passing the same kind of info as in the create? so that a user editing their entry can also have a drop down of GMs?
(i couldnt put this comment on that line of the file i think cus it wasnt being changed)

AnnieTran13 (approved):
Nice job! The field works great and as intended. 
We might need to make new issues or if you can address this in this PR:
- If the checkbox is unchecked (GM has no account), then the rating step should be skipped (right now it skips only if GM field is empty)
- When we edit the entry and add a GM who has an account, we should allow users to give a rating (right now there is no rating step in the edit drawer)

willyyhuang:
> Nice job! The field works great and as intended. We might need to make new issues or if you can address this in this PR:
> 
> * If the checkbox is unchecked (GM has no account), then the rating step should be skipped (right now it skips only if GM field is empty)
> * When we edit the entry and add a GM who has an account, we should allow users to give a rating (right now there is no rating step in the edit drawer)

I agree on the first one, but the 2nd one, im not sure if we allow ratings to be edited

m-triassi:
For that second point, we'll probably need to discuss with @DndMtl :) for now lets just leave it as is

m-triassi:
i'm going to say that code coverage here is good enough

PR #395, opened 2022-02-11
Trade index

### LINKED ISSUE #376: Trade Index
## Description

Users should be able to filter trades by item name, description, or rarity.

## Possible Implementation
Modify the trade index function to accept a filter and return filtered trades
Add the filterable trait to the trade model
Add a public static const to the trade model for trade status
Only open trades should be shown in the trades list

## Possible Extension
We may need to include a list of the current user's own trade offers in the index so that any given user can either
- view trade offers
- manage their own offers
both from the trade page.
Depends on the design.

bcobo341:
## Description
Filtered trades by requested_items, description, item name, item description, or item rarity.
Closes #376


app/Http/Controllers/TradeController.php
```diff
@@ -15,7 +16,28 @@ class TradeController extends Controller
      */
     public function index(Request $request)
     {
-        $trades = Trade::all();
+        $trades = Trade::where('status', 'open')
+                    ->has('items')->with('items')
+                    ->filtered($request->get('requested_items'))
+                    ->filtered($request->get('description'));
```
m-triassi, line 22: So the idea with the `Filterable` trait, is that it takes a query param and searches on all the fields at the same time so what i'd recommend here is just having 1 search parameter like this
```suggestion
                    ->filtered($request->get('search'))
```

app/Http/Controllers/TradeController.php
```diff
@@ -15,7 +16,28 @@ class TradeController extends Controller
...
+
+        if ($itemDescription = $request->get('item_description')) {
+            $trades = $trades->whereHas('items', function (Builder $q) use ($itemDescription) {
+                $q->where('description', 'like', "%{$itemDescription}%");
+            });
+        }
+
+        if ($itemRarity = $request->get('item_rarity')) {
+            $trades = $trades->whereHas('items', function (Builder $q) use ($itemRarity) {
+                $q->where('rarity', 'like', "%{$itemRarity}%");
```
m-triassi, line 38: the rarity of items are pretty set in stone, so in this case we dont need to do `like`, you can use a plain equals here:

```suggestion
                $q->where('rarity', $itemRarity);
```

app/Models/Trade.php
```diff
@@ -34,6 +36,10 @@ class Trade extends Model
         'character_id' => 'integer',
     ];
 
+    public const STATUS = ["open","closed"];
```
m-triassi, line 39: this should probably be 2 constants for less ambiguity 
```suggestion
    public const STATUS_OPEN = "open";
    public const STATUS_CLOSED = "closed";
```

tests/Feature/Http/Controllers/TradeControllerTest.php
```diff
@@ -35,7 +35,18 @@ public function setUp(): void
...
+            ->create();
+
+        $response = $this->get(route('trade.index', [
+            'requested_items' => $trades->first()->requested_items,
+            'description' => $trades->first()->description,
+            'item_name' => $trades->first()->items->first()->name,
+            'item_description' => $trades->first()->items->first()->description,
+            'item_rarity' => $trades->first()->items->first()->rarity,
+        ]));
+
```
m-triassi, line 49: there aren't any assertions in this test, would you be able to add some that check the trade items are there? Or maybe that the count of items returned in the trade the expected amount of items based off the applied filters?

tests/Feature/Http/Controllers/TradeControllerTest.php
```diff
@@ -35,7 +35,18 @@ public function setUp(): void
      */
     public function index_displays_view()
     {
-        $trades = Trade::factory()->count(3)->create();
+        $trades = Trade::factory(3)
+            ->has(Item::factory(3))
+            ->create();
```
m-triassi, line 40: This isnt the state a trade would be in; basically a trade will post an `item` and have offers in the `items` relation (note the s) so you'd need to define the specific relationships your assigning the items too, you can find how that works [here in the docs](https://laravel.com/docs/8.x/database-testing#has-many-relationships), note the second code block where they specify the relationship name

app/Http/Controllers/TradeController.php
```diff
@@ -15,7 +16,28 @@ class TradeController extends Controller
...
+        }
+
+        if ($itemDescription = $request->get('item_description')) {
+            $trades = $trades->whereHas('items', function (Builder $q) use ($itemDescription) {
+                $q->where('description', 'like', "%{$itemDescription}%");
+            });
+        }
+
+        if ($itemRarity = $request->get('item_rarity')) {
+            $trades = $trades->whereHas('items', function (Builder $q) use ($itemRarity) {
```
m-triassi, line 37: these `whereHas` should be on the `item` relation, since thats what is being offered 
```suggestion
            $trades = $trades->whereHas('item', function (Builder $q) use ($itemName) {
                $q->where('name', 'like', "%{$itemName}%");
            });
        }

        if ($itemDescription = $request->get('item_description')) {
            $trades = $trades->whereHas('item', function (Builder $q) use ($itemDescription) {
                $q->where('description', 'like', "%{$itemDescription}%");
            });
        }

        if ($itemRarity = $request->get('item_rarity')) {
            $trades = $trades->whereHas('item', function (Builder $q) use ($itemRarity) {
```

app/Http/Controllers/TradeController.php
```diff
@@ -15,7 +16,28 @@ class TradeController extends Controller
      */
     public function index(Request $request)
     {
-        $trades = Trade::all();
+        $trades = Trade::where('status', 'open')
+                    ->has('items')->with('items')
```
m-triassi, line 20: should be loading the item relation instead, and since the item relation will be essentially required, we wont need to check its existence
```suggestion
                    ->with('item')
```

m-triassi (changes requested):
its a good start! just needs some adjusting :)

tests/Feature/Http/Controllers/TradeControllerTest.php
```diff
@@ -35,13 +36,49 @@ public function setUp(): void
...
+                    ->where('id', $trades->first()->item->id)
+                    ->where('entry_id', $trades->first()->item->entry_id)
+                    ->where('character_id', $trades->first()->item->character_id)
+                    ->where('name', $trades->first()->item->name)
+                    ->where('author_id', $trades->first()->item->author_id)
+                    ->where('rarity', $trades->first()->item->rarity)
+                    ->where('tier', $trades->first()->item->tier)
+                    ->where('description', $trades->first()->item->description)
+                    ->etc()
+                )
```
m-triassi, line 78: oh wow, nice, this is really proper testing with inertia responses; good work!

m-triassi (approved):
This looks pretty excellent to me!

PR #396, opened 2022-02-11
Empty Reward Choice and Assigned Character Fields When Editing a DM Entry

### LINKED ISSUE #343: Saving after editing a DM entry removes the reward choice and assigned character
## Description

After creating an entry with a reward choice selected and a character assigned, when we edit it, those fields will come back empty on the edit drawer. If you save your edits without editing those fields, the reward choice selected and the character assigned to that entry will be removed.

## Expected Behavior

When editing a DM entry, the edit drawer must show the reward choice selected and assigned character (if any) of that entry.

## Actual Behavior

When editing a DM entry, the edit drawer shows an empty field for the reward choice selected and assigned character (if any) of that entry. After saving without filling those fields, the reward choice selected and assigned character are removed.

## Steps to Reproduce

1. Go to the character's page, select DM entry.
2. Create a DM entry with a reward choice selected and a character assigned.
3. Edit the DM entry without touching the reward choice and the character assigned fields.
4. Click save.
5. The entry no longer has the reward choice and the character assigned it was initially selected/assigned

## Possible Fix

# Other

AnnieTran13:
closes #343 

Reward choice and assigned character in a DM entry (if any) shows when editing the entry (used to be blank). The changes made affected the format of the rewards column in EntryTable and DMEntryTable (used to be lowercase with underscores) but it was fixed in this PR as well.


app/Http/Requests/EntryUpdateRequest.php
```diff
@@ -49,7 +50,7 @@ public function rules()
             'gp' => ['sometimes', 'numeric', 'between:-999999999999999999999999999999.99,999999999999999999999999999999.99'],
             'downtime' => ['sometimes', 'integer'],
             'notes' => ['nullable', 'string'],
-            'items' => ['sometimes', 'array'],
+            'items' => ['sometimes', 'array', $requiredChoice],
```
pascwhale, line 53: Hey @m-triassi, could you explain the `$requiredChoice` expression? The way I read it is that an array of items is required if the user's chose `Magic Items` as a reward and the items in the request were empty. Shouldn't the `items` array not be empty?

pascwhale (changes requested):
Everything seems to work well, just had a few questions regarding some backend code.

willyyhuang (approved):

app/Http/Requests/EntryUpdateRequest.php
m-triassi: What you're saying is mostly correct, it just stipulates that there should be _something_ for magic items if the choice is "magic_items", in the `$requiredChoice` we check both what the value of choice is **and** that the items array isn't empty. 

So to summerize, the `$requiredChoice` is --> is choice == "magic_item"? yes. is `items` not empty? yes --> you're good, if not then validation fails.

app/Http/Requests/EntryUpdateRequest.php
```diff
@@ -33,11 +33,12 @@ public function rules()
     {
         $rarities = implode(",", Item::RARITY);
         $requiredIf = Rule::requiredIf(!empty($this->get('items')));
+        $requiredChoice = Rule::requiredIf($this->get('choice') == "magic_item" && empty($this->get('items')));
 
         return [
             'adventure.id' => ['required', 'integer', 'exists:adventures,id'],
             'campaign_id' => ['sometimes', 'integer', 'exists:campaigns,id'],
-            'character_id' => ['nullable', 'integer', 'exists:characters,id'],
+            'character_id' => ['nullable', 'integer', 'exists:characters,id', 'required_with:choice'],
```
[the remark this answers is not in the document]
m-triassi: yeah its super handy

m-triassi:
@AnnieTran13 Would you be able to just resolve the merge conflicts here? otherwise i think we're good to merge this!

PR #397, opened 2022-02-13
Add warning related to character import

### LINKED ISSUE #393: Imported item tier should not be 0
## Description

imported item tier should not be 0, instead front-end should add a reminder for user to set the item tier manually

## Possible Implementation

add a button or tag to remind user to edit the item tier if the tier is = 0, onClick should open the edit drawer

willyyhuang:
Closes #393 

Added:
- Button to edit when tier is not set (0) / adventure not set
- Help text for character import
- New day formatting
- Update theme colors


willyyhuang:
Ignore the complexity warning, this component needs to cover multiple edge cases, therefore we need the multiple return statements

m-triassi:
The diff coverage is only 70% but it looks like it wouldn't change the total coverage here, so if a front-end approve the code itself, you can go ahead and merge this

AnnieTran13 (approved):
- Button to set tier in the items table is clear and easy to use. The orange-yellow color for it is nice too
- Help text is very clear. I like the coloring given so that the user can't miss it.
- Day format works as intended
-Theme color update works as intended

PR #398, opened 2022-02-13
Change Entry Update to no longer delete existing items

### LINKED ISSUE #381: Refactor Entry Item Editing
## Description
When an entry that has items is edited, the items are deleted and remade. This could be a source of problems with item trading, and should be refactored so that items are not deleted when the entry is edited.

## Possible Implementation
One idea is to simply prevent an item or entry to be edited if the item has been traded. The request would return to the entry edit page with an error response code that can be displayed in some way by front end.

m-triassi:
## Description
Closes #381
Changes the Entry update endpoint account for existing items. Achieves this by loading items that already exist and opting to update them rather than deleting all items and then re-creating them.

### DIFF
--- a/app/Actions/CreateEntryItems.php
+++ b/app/Actions/CreateEntryItems.php
@@ -21,8 +21,8 @@ class CreateEntryItems
     {
         $out = collect();
         $items = collect($items)->sortBy('id');
-        $existing = Item::whereIn('id', $items->pluck('id')->filter())
-            ->where('character_id', $entry->character_id)
+        $existing = $entry->items()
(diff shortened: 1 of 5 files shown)


willyyhuang:
Cypress test cases failed

IanjhPhillips (approved):
Tested locally by updating an entry to change items, or add new ones, it all behaved as expected.
I also had the admin panel open in another tab and was watching th items list which (i assume) is sorted by id, and it didnt change which indicated to me that the items werent being deleted.

Nicely done!

PR #400, opened 2022-02-13
Split cypress pipeline

### LINKED ISSUE #389: Split up CI Pipeline
## Description

The current pipeline takes about +10 minutes to run. It would benefit from some parallelization. The majority of the run time is taken up by the cypress E2E testing step.

## Possible Implementation

Since the jest test suite and the PHP unit test suites need  to be run together (to post reports to code climate), the only other step would be to separate out the Cypress step into its own pipeline. Also add proper caching to the front-end `node_modules` + build assets

willyyhuang:
Closes #389 
Split cypress pipeline due to long sequential pipeline build time

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -43,6 +43,12 @@ jobs:
         run: |
           composer install -n --no-ansi --no-interaction --no-scripts --no-progress --prefer-dist
 
+      - name: Cache FE dependencies
+        uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+
       - name: Setup Node
         uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-suite.yml
```diff
@@ -71,6 +71,37 @@ jobs:
             tests/coverage.xml:clover
             coverage/clover.xml:clover
 
+  Cypress:
+    runs-on: ubuntu-latest
+    services:
```
willyyhuang, line 76: i'm not sure if this is needed for `cypress` tbh @m-triassi

m-triassi: if you mean the `services` block, then thats definitely required, the tests need to be able to read and write data to a database. if you mean the `runs-on` line, that's also required, we need to define what OS to run the pipeline on

willyyhuang: yeah i was just wondering if the service block was needed

pascwhale (approved):
Everything looks good, it seems like you copied the right steps for the new pipeline. I'm happy that this will lessen the delay of our checks.

m-triassi (approved):
Agree'd this is a good stop gap until we find a better way to speed up the cypress suite in general

.github/workflows/test-suite.yml
m-triassi: i see, yeah services is definitely needed

PR #401, opened 2022-02-14
Missing design feedbacks

willyyhuang:


pascwhale (approved):
Everything looks good. The hints somehow sound friendlier. The alert looks a lot better too.

PR #402, opened 2022-02-18
User profile

### LINKED ISSUE #383: User profile
## Description
Add user profile, should have:
- change password
- forget password
- language ? translation??

willyyhuang:
Closes #383 
Added user profile form, letting user to delete or edit their account


resources/js/Pages/Profile.tsx
```diff
@@ -0,0 +1,191 @@
...
+        put,
+        reset,
+        wasSuccessful,
+        processing,
+        delete: destroy,
+    } = useForm<UserEditDataType>(USER_EDIT_INITIAL_VALUE)
+    const [isDeleteModalVisible, setIsDeleteModalVisible] = useState<boolean>(false)
+    const [isSuccess, setIsSuccess] = useState<boolean>(wasSuccessful)
+
+    useEffect(() => {
```
willyyhuang, line 50: in case you don't understand what this `useEffect` hook does, basically every time `wasSuccessful` is changed, this hook will set the value of `isSuccess` to `wasSuccessful` and `isSuccess` is the visible state of the `Snackbar` which displays a success message and will automatically change `isSuccess` to `false` after 3 seconds.

willyyhuang: for future reference, I will refactor this and extract to a method to display notifications most likely.

pascwhale (approved):
Everything works really smoothly @willyyhuang! Two things I would maybe change that are entirely optional: more padding between a section title and its input, and disabling (greying out) `RESET` and `SAVE` buttons when no changes are made. Otherwise, great work!

PR #403, opened 2022-02-18
Trade Fulfillment

### LINKED ISSUE #378: Trade Editing and Fulfillment
## Description
Modify the trade update endpoint to allow users to modify their trade info.
If a user modifies their trade after a number of offers have already been made, the offers must be removed from the trade.

Additionally, the edit endpoint should accept an item id and a status to indicate that a trade offer has been accepted by the user who originally listed the trade. The trade must be then set to closed and the items should be swapped. 

## Possible Implementation
It might be a good idea to create an action for swapping two items between characters.

IanjhPhillips:
## Description
Closes #378 
Creates an action to swap items between two characters based on a trade
Creates an endpoint to handle closing a trade which calls the above action


app/Actions/FulfillTrade.php
```diff
@@ -0,0 +1,63 @@
...
+class FulfillTrade
+{
+    use AsAction;
+
+    /**
+     * Expect an array of item details, create each item and attach them to the appropriate relation objects
+     *
+     * @param Entry $entry
+     * @param array $items
+     */
```
m-triassi, line 23: oops your copy paste is showing... update this doc block please. also you can use `php artisan make:action ActionName` in the future instead

app/Actions/FulfillTrade.php
```diff
@@ -0,0 +1,63 @@
...
+        $tradeCharacter = $trade->character;
+        $tradeItem = $trade->item;
+        $offerCharacter = $offeredItem->character;
+
+        $currentUser = Auth::user();
+        $offerUser = $offerCharacter->user;
+
+        if (!$currentUser->characters->contains($tradeCharacter) || $trade->status == Trade::STATUS_CLOSED) {
+            throw new TradeException("The trade could not be fulfilled.");
+        }
```
m-triassi, line 35: This is good, but these two errors represent 2 different status codes, the first clause is one of permission, while the other is of timing. Custom exception classes can be pretty specific, usually. Something like this might be good: 
```suggestion
        if (!$currentUser->characters->contains($tradeCharacter)) {
            abort(403);
        }
        
        if ($trade->status == Trade::STATUS_CLOSED)) {
            throw new TradeClosedException();
        }
```
then the `TradeClosedException` class would specify the message and status code (400) in the construct method, which you'd override from the parent. you could also, instead of calling `abort(403)`, make a second exception class, something like `TradeNotAllowedException` that would throw a 403 for any dis-allowed trade actions, again setting the message and code in the construct.

app/Actions/FulfillTrade.php
```diff
@@ -0,0 +1,63 @@
...
+
+        $offeredItem->character()->associate($tradeCharacter->id)->save();
+        $tradeItem->character()->associate($offerCharacter->id)->save();
+
+        $entryData = [
+            'date_played' => now(),
+            'type' => Entry::TYPE_DOWNTIME,
+            'levels' => 0,
+            'notes' => "Trade between ".$tradeCharacter->name." and ".$offerCharacter->name
+        ];
```
m-triassi, line 48: This should also subtract 5 downtime days

app/Actions/FulfillTrade.php
```diff
@@ -0,0 +1,63 @@
...
+            'character_id' => $tradeCharacter->id]));
+        $offeredItem->entry()->associate($tradeCharacterEntry)->save();
+
+        $offerCharacterEntry = Entry::create(array_merge($entryData, [
+            'user_id' => $offerUser->id,
+            'character_id' => $offerCharacter->id]));
+        $tradeItem->entry()->associate($offerCharacterEntry)->save();
+
+        $trade->status = Trade::STATUS_CLOSED;
+        $trade->save();
```
m-triassi, line 61: we may want to add a step where we delete all the other pending offers, no?

m-triassi (changes requested):
Really good stuff, just a few things that need attention

app/Actions/FulfillTrade.php
IanjhPhillips: oopsie

app/Actions/FulfillTrade.php
IanjhPhillips: ah good idea, ill look into this

app/Actions/FulfillTrade.php
IanjhPhillips: good idea, ill add that

app/Actions/FulfillTrade.php
IanjhPhillips: done

app/Actions/FulfillTrade.php
IanjhPhillips: done, please check that its okay

app/Actions/FulfillTrade.php
IanjhPhillips: done

app/Actions/FulfillTrade.php
IanjhPhillips: fixed.

app/Actions/FulfillTrade.php
```diff
@@ -0,0 +1,76 @@
...
+            'user_id' => $offerUser->id,
+            'character_id' => $offerCharacter->id]));
+        $tradeItem->entry()->associate($offerCharacterEntry)->save();
+
+        foreach ($trade->offers as $offer) {
+            if ($offer->id == $offeredItem->id) {
+                continue;
+            }
+            $trade->offers()->detach($offer->id);
+        }
```
m-triassi, line 71: you can just call
```suggestion
       $trade->offers()->sync([])
```
that'll detach them all in 1 go

m-triassi: we can detach all offers since the trade is closing anyway

m-triassi (approved):
Yeah this looks pretty good, one last optional comment for you

PR #410, opened 2022-02-20
Trade Offer Creation and Updates

JasonGadoury:
Edit and create functions tests failing. The issue is most likely in the way the routes were created or the way the request parameters are passed. Both return views at the moment, assuming this will be replaced by Inertia::render() calls. Any suggestions  appreciated for how to fix.

### DIFF
--- a/app/Http/Controllers/TradeOfferController.php
+++ b/app/Http/Controllers/TradeOfferController.php
@@ -1,0 +1,122 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Http\Requests\TradeStoreRequest;
+use App\Http\Requests\TradeUpdateRequest;
+use App\Models\Character;
+use App\Models\Item;
+use App\Models\Trade;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+use Inertia\Inertia;
+use function PHPUnit\Framework\isNull;
+
+class TradeOfferController extends Controller
+{
+    /**
+     * Retrieve character's items that are valid to trade
+     *
+     * @param \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function create(Request $request)
+    {
+        //REQUEST DOES NOT HAVE TRADE_ID NOR CHARACTER_ID . VALIDATION FAILS
+        $offerData = $request->validate([
+            'trade_id' => 'required|exists:trades,id|integer',
+            'character_id' => "required|exists:characters,id|integer",
+        ]);
+
+        $character = Character::where('user_id', Auth::id())
+            ->findOrFail($offerData['character_id']);
+        $trade = Trade::findOrFail($offerData['trade_id']);
+        $tradeItem = $trade->item()->get();
+
+        $validItems = $character->items()->where('tier', $tradeItem->tier);
+
+        return view('offer.create', compact('validItems'));
+    }
+
+    /**
+     * Retrieve Trade Offer to edit
+     *
+     * @param \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function edit(Request $request)
+    {
+        //rrequest params passed in "headers" instead of request parameters...
+
+        $offerData = $request->validate([
+            'trade_id' => 'required|exists:trades,id|integer',
+            'item_id' => 'required|exists:items,id|integer',
+        ]);
+
+        $trade = Trade::findOrFail($offerData['trade_id']);
+        $item = Item::findOrFail($offerData['item_id']);
+
+        return view('offer.edit', compact('item', 'trade'));
+    }
+
+    /**
+     *store trade offer
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function store(Request $request)
+    {
+        $offerData = $request->validate([
+            'item_id' => 'required|exists:items,id|integer',
+            'trade_id' => 'required|exists:trades,id|integer'
+        ]);
+
+        //findOrFail redundant?
+        $trade = Trade::findOrFail($offerData['trade_id']);
+        $dataItem = Item::findOrFail($offerData['item_id']);
+        $offerItem = ($dataItem->tier == $trade->item()->get()[0]->tier) ? $dataItem : null;
+
+
+        // Attach offered item that meets rarity criteria
+        if (!is_null($offerItem)) {
+            $trade->items()->attach($offerItem);
+        }
+
+        return redirect()->back();
+    }
+
+    /**
+     * Update existing trade offer
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \App\Models\Trade $trade
+     * @param \App\Models\Item $item
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function update(Request $request)
+    {
+        //TRADE & ITEM EMPTY, might be problem with route?
+        $offerData = $request->validate([
+            'item_id' => 'required|exists:items,id|integer',
+            'replace_id' => 'required|exists:items,id|integer',
+            'trade_id' => 'required|exists:trades,id|integer'
+        ]);
+
+        //item to replace existing offer
+        $oldOffer = Item::findOrFail($offerData['replace_id']);
+        $newOffer = Item::findOrFail($offerData['item_id']);
+        $trade = Trade::findOrFail($offerData['trade_id']);
+
+
+        //does not work since $trade empty
+        $newOffer = ($newOffer->tier == $trade->item()->get()[0]->tier) ? $newOffer : null;
+
+        if (!is_null($newOffer)) {
+            $trade->items()->detach($oldOffer);
+            $trade->items()->attach($newOffer);
+        }
+
+        return redirect()->back();
+    }
+}
--- a/routes/web.php
+++ b/routes/web.php
@@ -4,6 +4,7 @@ use Illuminate\Foundation\Application;
 use Illuminate\Support\Facades\Route;
 use Inertia\Inertia;
 use Illuminate\Support\Str;
+use App\Http\Controllers\TradeOfferController;
 
 require __DIR__ . '/auth.php';
 
@@ -53,6 +54,17 @@ Route::middleware(['auth', 'throttle'])->group(function () {
     Route::resource('item', App\Http\Controllers\ItemController::class);
 
     Route::resource('trade', App\Http\Controllers\TradeController::class);
+
+    Route::put('/offer/update', [App\Http\Controllers\TradeOfferController::class, 'update'])
+        ->name("offer.update");
+
+    Route::get('/offer/edit', [App\Http\Controllers\TradeOfferController::class, 'edit'])
+        ->name("offer.edit");
+
+    Route::get('/offer/create', [App\Http\Controllers\TradeOfferController::class, 'create'])
+        ->name("offer.create");
+
+    Route::resource('offer', App\Http\Controllers\TradeOfferController::class)->except('update', 'edit', 'create');
 
     Route::resource('event', App\Http\Controllers\EventController::class);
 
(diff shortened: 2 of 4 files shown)


JasonGadoury:
Seems like items() relationship in Trade is now offers() in main. I left it as items() so I'll refactor all items() calls to offers()

app/Http/Controllers/TradeOfferController.php
```diff
@@ -0,0 +1,122 @@
...
+
+        $character = Character::where('user_id', Auth::id())
+            ->findOrFail($offerData['character_id']);
+        $trade = Trade::findOrFail($offerData['trade_id']);
+        $tradeItem = $trade->item()->get();
+
+        $validItems = $character->items()->where('tier', $tradeItem->tier);
+
+        return view('offer.create', compact('validItems'));
+    }
```
m-triassi, line 39: you can just do this
```suggestion
    public function create(Request $request)
    {
        $offerData = $request->validate([
            'trade_id' => 'required|exists:trades,id|integer',
        ]);
        
        $trade = Trade::with('item')->findOrFail($offerData['trade_id']);
        $validItems = Auth::user()->items()->where('rarity', $trade->item->rarity)
        
        // will be replaced when page component is created...
        return view('offer.create', compact('validItems', 'trade'));
    }
```

app/Http/Controllers/TradeOfferController.php
```diff
@@ -0,0 +1,122 @@
...
+    public function store(Request $request)
+    {
+        $offerData = $request->validate([
+            'item_id' => 'required|exists:items,id|integer',
+            'trade_id' => 'required|exists:trades,id|integer'
+        ]);
+
+        //findOrFail redundant?
+        $trade = Trade::findOrFail($offerData['trade_id']);
+        $dataItem = Item::findOrFail($offerData['item_id']);
```
m-triassi, line 77: you're right, it is redundant, but it doesn't add much overhead, you're call if you want to keep it or not

app/Http/Controllers/TradeOfferController.php
```diff
@@ -0,0 +1,122 @@
...
+        $offerItem = ($dataItem->tier == $trade->item()->get()[0]->tier) ? $dataItem : null;
```
m-triassi, line 78: if the offered item is the wrong rarity you should actually raise an exception or return back with an error of some sort.

app/Models/Trade.php
```diff
@@ -42,7 +42,7 @@ class Trade extends Model
     protected $filterableFields = ['requested_items', 'description'];
 
 
-    public function offers()
+    public function items()
```
m-triassi, line 45: why did you change this back?

tests/Feature/Http/Controllers/TradeOfferControllerTest.php
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * @test
+     */
+    public function edit_displays_offer_view()
+    {
+        $trade = Trade::factory()->create([
+        ]);
+        $item = Item::factory()->create([
+        ]);
```
m-triassi, line 38: please watch your code formatting

m-triassi (changes requested):
Some stuff to address here

routes/web.php
```diff
@@ -54,6 +55,17 @@
...
+    Route::put('/offer/update', [App\Http\Controllers\TradeOfferController::class, 'update'])
+        ->name("offer.update");
+
+    Route::get('/offer/edit', [App\Http\Controllers\TradeOfferController::class, 'edit'])
+        ->name("offer.edit");
+
+    Route::get('/offer/create', [App\Http\Controllers\TradeOfferController::class, 'create'])
+        ->name("offer.create");
+
+    Route::resource('offer', App\Http\Controllers\TradeOfferController::class)->except('update', 'edit', 'create');
```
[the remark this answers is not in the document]
JasonGadoury: I did this because some of the default routes (before I added ->except(...)) expected parameters even though I didn't necessarily have any to pass. I left this because it works fine for store and I was thinking we might define more controller functions later on. I suppose I can just remove this and manually add a store route for the moment.

app/Http/Controllers/TradeOfferController.php
```diff
@@ -0,0 +1,122 @@
...
+{
+    /**
+     * Retrieve character's items that are valid to trade
+     *
+     * @param \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function create(Request $request)
+    {
+        //REQUEST DOES NOT HAVE TRADE_ID NOR CHARACTER_ID . VALIDATION FAILS
```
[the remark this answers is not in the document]
JasonGadoury: Did this for debugging purposes, will delete it entirely once fixed

app/Http/Controllers/TradeOfferController.php
```diff
@@ -0,0 +1,122 @@
...
+
+        //findOrFail redundant?
+        $trade = Trade::findOrFail($offerData['trade_id']);
+        $dataItem = Item::findOrFail($offerData['item_id']);
+        $offerItem = ($dataItem->tier == $trade->item()->get()[0]->tier) ? $dataItem : null;
+
+
+        // Attach offered item that meets rarity criteria
+        if (!is_null($offerItem)) {
+            $trade->items()->attach($offerItem);
```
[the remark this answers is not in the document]
JasonGadoury: Yeah I noticed that. from code climate Will do

app/Http/Controllers/TradeOfferController.php
```diff
@@ -0,0 +1,122 @@
...
+    public function create(Request $request)
+    {
+        //REQUEST DOES NOT HAVE TRADE_ID NOR CHARACTER_ID . VALIDATION FAILS
+        $offerData = $request->validate([
+            'trade_id' => 'required|exists:trades,id|integer',
+            'character_id' => "required|exists:characters,id|integer",
+        ]);
+
+        $character = Character::where('user_id', Auth::id())
+            ->findOrFail($offerData['character_id']);
```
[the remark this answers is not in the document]
JasonGadoury: Got it, thanks

tests/Feature/Http/Controllers/TradeOfferControllerTest.php
JasonGadoury: I had provided specific values to the trade and item's fields for debugging purposes hence the square brackets and line breaks. This test was left incomplete because I couldn't figure out why  `$response = $this->get(route('offer.edit'), ...`  was not successfully passing the request parameters to the controller. I'm removing the update and edit methods so this test is getting deleted.

app/Models/Trade.php
JasonGadoury: I must have merged main into my branch earlier on so I used items() all throughout the feature. I'll refactor all of it before merging.

app/Http/Controllers/TradeOfferController.php
JasonGadoury: @m-triassi is it tier or rarity that we're looking for?

JasonGadoury: I modified this suggestion to accept a trade parameter instead. Adjusted route and tests are now passing

tests/Feature/Http/Controllers/TradeOfferControllerTest.php
```diff
@@ -0,0 +1,122 @@
...
+        $item = Item::factory()->create([
+            'rarity' => 'rare',
+            'id' => 234,
+            'description' => "listed"
+        ]);
+
+        $trade = Trade::factory()->create([
+            'id' => 456,
+            'item_id' => $item->id
+        ]);
```
m-triassi, line 67: you don't need to / shouldn't explicitly set the `id` of the model here, there's a change for a collision. When the factory returns from `create()` the id will have been generated for you.
```suggestion
            'rarity' => 'rare',
            'description' => "listed"
        ]);

        $trade = Trade::factory()->create([
            'item_id' => $item->id
        ]);
```

tests/Feature/Http/Controllers/TradeOfferControllerTest.php
```diff
@@ -0,0 +1,122 @@
...
+    public function destroy_detaches_offer()
+    {
+        $offer = Item::factory()->create([
+            'id' => 333,
+            'description' => "test item"
+        ]);
+        $trade = Trade::factory()->create();
+        $trade->offers()->attach($offer);
+
+        $response = $this->from('/')->delete(route('offer.destroy', [$offer, $trade]));
```
m-triassi, line 117: same here `from(...)` not need
```suggestion
        $response = $this->delete(route('offer.destroy', [$offer, $trade]));
```

app/Http/Controllers/TradeOfferController.php
```diff
@@ -0,0 +1,70 @@
...
+     * @param \App\Models\Item $item
+     * @param \App\Models\Trade $trade
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function destroy(Request $request, Item $offer, Trade $trade)
+    {
+        $trade->offers()->detach($offer);
+
+        return redirect()->back();
+    }
```
m-triassi, line 69: just noticed that there isn't any validation that checks if a user owns a trade before they deleted it, you might want to add that

tests/Feature/Http/Controllers/TradeOfferControllerTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * @test
+     */
+    public function destroy_detaches_offer()
+    {
+        $offer = Item::factory()->create([
+            'id' => 333,
+            'description' => "test item"
+        ]);
```
m-triassi, line 113: when you add the authorization check to the `destroy()` function, you'll want the owner of this trade to be owned by a character that `$this->user` owns...

m-triassi (changes requested):
I think you're really close, just a few more things!

tests/Feature/Http/Controllers/TradeOfferControllerTest.php
JasonGadoury: This was done for debugging purposes to make it easier when using dd() to spot if the relationships contained the intended objects. But I agree, I should probably remove it after confirmation.

tests/Feature/Http/Controllers/TradeOfferControllerTest.php
```diff
@@ -0,0 +1,122 @@
...
+            'description' => "listed"
+        ]);
+
+        $trade = Trade::factory()->create([
+            'id' => 456,
+            'item_id' => $item->id
+        ]);
+        //If item_id not explicitly set as in line 87, the item_id increments by 1
+        //after during store request... below line not sufficient
+        //$trade->item()->associate($item);
```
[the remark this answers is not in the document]
JasonGadoury: Ahh that explains it. This is why I resorted to setting specific ids to the trade and item. I was attempting to associate them but when I checked the $trade->item->id it was giving me whatever the item's id was + 1 and a different rarity. Thanks for the reminder!

tests/Feature/Http/Controllers/TradeOfferControllerTest.php
```diff
@@ -0,0 +1,122 @@
...
+            'trade_id' => $trade->id,
+            'item_id' => $offer->id
+        ]);
+
+        $response->assertRedirect('/');
+
+        $response = $this->from('/')->post(route('offer.store'), [
+            'trade_id' => $trade->id,
+            'item_id' => $invalidOffer->id
+        ]);
```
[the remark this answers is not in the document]
JasonGadoury: What is an alternative to checking where redirect()->back() will take us in that case? I'm not exactly sure what the previous uri (i.e. the ->back()) is without explicitly setting it.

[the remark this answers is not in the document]
m-triassi: You can just check that you got redirected, `$response->assertRedirect()`, since there isn't a specific destination, your tests can just assume that the underlying framework is doing its job properly

m-triassi (approved):
This seems ready! nice work! :)

PR #413, opened 2022-02-21
Campaign Join Form and Modal

### LINKED ISSUE #387: Campaign join form
## Description

See #251 for narrative.
There should be a form that allows users to join a campaign. It should have a field where the user can enter a campaign join code. This field can be prefilled if the user uses a join link. The page must show the characters a player can join as. It should also present the option to join as a GM instead. If a user chooses to join as a GM then they should not be able to choose a character

AnnieTran13:
closes #387 

Campaign join button opens a model where user inputs their invite code. After that they are redirected to the join form where they need to assign characters to the campaign.

Incomplete because blocked by Campaign Index Page


resources/js/Components/CampaignJoinModal/CampaignJoinModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+    message: string
+}
+
+type CampaignJoinModalDataType = {
+    code: string | undefined
+}
+
+const CampaignJoinModal = ({open, onClose, message}: CampaignJoinModalPropType) => {
+    const CAMPAIGN_JOIN_MODAL_FORM_INITIAL_VALUE: CampaignJoinModalDataType = {
+        code: undefined,
```
willyyhuang, line 42: don't use constant here

resources/js/Components/CampaignJoinModal/CampaignJoinModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+    const {data, setData, errors, clearErrors, post} = useForm(
```
willyyhuang, line 44: type useForm

willyyhuang (approved):

PR #415, opened 2022-02-21
Event Reports Cypress Test

pascwhale:
## Description
Closes #322
Write E2E Cypress tests for #247 Event Reports user story acceptance criteria.

## Acceptance Criteria
- Export a report for an event

PR #422, opened 2022-02-21
Fix missing Cypress Test

m-triassi:
## Description
Adds back missing cypress test from bad merge + revert in #415


willyyhuang (approved):

pascwhale (approved):
Yup, this is the missing file for the Event Reports Cypress tests. Thank you!

PR #427, opened 2022-02-21
Loading button

### LINKED ISSUE #407: Loading button
## Description

Add a loading state to buttons since Inertia supports `processing` in `useForm`

## Possible Implementation

Create a new atom component for button, add the prop: `loading`
Display https://mui.com/components/progress/ instead of regular text when processing is true

There is an implementation on it in `Profile.tsx` right now

willyyhuang:
Closes #407 
Added loading button component which shows circular motion when making requests
This also fixes the user name being too long on resized window

Notable place to test:
character import

### DIFF
--- a/resources/js/Components/Atom/Button.tsx
+++ b/resources/js/Components/Atom/Button.tsx
@@ -1,0 +1,37 @@
+import {
+    Button as MuiButton,
+    ButtonProps as MuiButtonProps,
+    CircularProgress,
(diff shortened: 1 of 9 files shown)


luigibesani (approved):
As hinted in the PR description, the character import's request is long enough to trigger the loading animation on the final button. Looks great. Good work @willyyhuang!

Issue #423, opened 2022-02-21
Edit error handling


willyyhuang:
## Description

Currently, if an edit request is posted, regardless of success or failure, the edit drawer will close.
We need better error handling on this.

## Expected Behavior

If request fails: display error message + don't close the drawer
If request passes: close drawer, preserve scrolling

## Actual Behavior

Regardless of success/failure, edit drawer will close

## Steps to Reproduce

1.  Import a character (Grod for example)
2.  Edit Grod's entry
3.  Modify notes
4. Request fails, drawer closes

## Possible Fix

Better handling on edit for backend redirecting or fix it in front-end somehow (Promise object)
This issue happens because `errors` prop gets updated after the `put` request, therefore the `form` thinks its successful since `errors` is an empty object (`{}`)

# Other

willyyhuang:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Session-Tome-613674a263226c0015d44c06/issues/DnD-Montreal/session-tome/423?planning-poker) with ZenHub @AnnieTran13 @luigibesani @pascwhale

PR #441, opened 2022-02-22
English/French translations + edit drawer error handling

### LINKED ISSUE #385: ENGLISH/FRENCH Translation
## Description
Add translation linked to user account
if user chose French as preferred language, show all the strings in french

## Possible Implementation
react i18n

willyyhuang:
Closes #385 , #423

Added react-i18next to handle english/french translations
Edit drawer should not close on failure as well now.

Usage of react-i18next is
- import `useTranslation` hook (since we are using functional components, hooks work the best)
- use the t function to find the specific string in the translation files


resources/js/Components/Detail/ItemDetailBox.tsx
```diff
@@ -32,101 +33,75 @@ type ItemDetailBoxPropType = {
...
-                        </Button>
-                        <Button
-                            variant='contained'
-                            startIcon={<CompareArrowsIcon fontSize='small' />}>
-                            TRADE
-                        </Button>
-                    </Stack>
+                    <StyledTypography>{t('itemDetail.description')}</StyledTypography>
+                    <Typography>
+                        {item.description === '""' ? 'No description' : item.description}
```
willyyhuang, line 98: To fix

willyyhuang:
Diff-coverage is 89.9% and total coverage increased by 0.3%, the reason why diff-coverage can not be achieved to 90% is because there are conditions where we can't test in unit tests now, such as autocomplete fields and `useEffect` hooks.

luigibesani (approved):
Excellent work! You really were thorough with the translations; it seems that everything is covered.
The implementation of the translation framework is very clean as well.
We should discuss about the presence of the language selection drop down in the front page.

PR #454, opened 2022-02-26
Add new event accessors

### LINKED ISSUE #435: Add extra Event model accessors
## Description

There are a number of pieces of data that will need to be displayed on both the event index, as well as the event details page.
This data includes: 
- **Event** date (single date or range)
- Event Capacity (total seats available)
- Amount of registered users (# of session registrations)
- Total spots available for an event (sum of all open seats)

See images for usage / location

image
image

## Possible Implementation

Add accessors to event for each of these items. add the added accessors to the `$appends` property on the `Event` Model.

- **Event** date --> pluck all start times from event sessions sorted by ascending, if it spans multiple days, generate a date range (Jan 1 - Jan 3), otherwise just return the day the sessions are on.
- Event Capacity --> pluck seats from event sessions, take sum
- Total spots available for event --> pluck open_seats accessor, take sum
- Amount of registered users --> subtract event capacity and available seats values

## Out of scope

N/a

N/a

m-triassi:
# Description
Closes #435 

Adds accessors to the `Event` model that expose some statistics about the event. Statistics such as: Open seats, taken seats, seats left, and date range of the event.

## How to test
1. Via tinker Create an event (or find one in the seeded data)
2. Attach some sessions to that event (or ensure the seeded version has some)
3. Register some users / characters to some sessions on that event
4. Check the values of `$event->total_seats`, `$event->seats_taken`, `$event->seats_left`
5. Ensure that the above values based off the amount of seats available (sum of seats on sessions, count of registered users, etc)
6. Check the value of  `$event->scheduled_dates` it should be an array with 2 elements the first is the start, the second the end (times of first and last session)


app/Models/Event.php
```diff
@@ -47,4 +54,32 @@ public function league()
...
+    public function getScheduledDatesAttribute()
+    {
+        $dates = $this->sessions()
+            ->orderBy('start_time')
+            ->pluck('start_time')
+            ->unique();
+
+        return [
+            $dates->first(),
+            $dates->last()
```
IanjhPhillips, line 67: Why are we only returning the first and last session start dates? Since the dates are unique is there any reason to not just return them all? Im sure there is some situation where the list of all start dates would be useful.

IanjhPhillips (approved):
Nicely done, everything is clear in how it works and does what it says it does.

Only had on comment about the output of one of the accessors, otherwise everything is good!

app/Models/Event.php
m-triassi: its mostly for brevity, this effectively represents the date range for the event, we _could_ return the whole collection, sure, but that's unnecessary for the use case of this accessor

IanjhPhillips: okay sounds good. in that case i have nothing else to add or change

PR #455, opened 2022-02-27
Session Overlap Check

### LINKED ISSUE #431: Prevent Session timeslot overlap registration
## Description

Prevent a user from registering for a particular session if they're already registered for a session with a time slot that overlaps.
Add End time for session

## Possible Implementation

Do this by checking the start / end times of each session the user is registered to. The easiest way would be to add a `hasManyThrough` relation to `User` with the `Session` model, then iterate over that list and check for overlap in the `EventRegistrationController`.

## Out of scope

**This is not required**
Adding "slop" time would be nice, aka if 2 sessions overlap by 5 minutes, we could potentially allow that.

Should be global, across all sessions a user is registered for.

IanjhPhillips:
## Description
Closes #431
When a user attempts to register for an event session, the system will check if that session overlaps with any sessions that the user is registered for as a player or GM.

### DIFF
--- a/app/Http/Controllers/Admin/SessionCrudController.php
+++ b/app/Http/Controllers/Admin/SessionCrudController.php
@@ -44,6 +44,7 @@ class SessionCrudController extends CrudController
         CRUD::column('dungeonMaster');
         CRUD::column('table');
         CRUD::column('start_time');
+        CRUD::column('end_time');
     }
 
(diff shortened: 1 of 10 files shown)


m-triassi (approved):
This looks pretty stellar, i cant really think of anything that needs changing!

PR #457, opened 2022-02-27
Return to step 1 when there is error for edit drawers

willyyhuang:
Fix a bug where edit drawer does not return to step 1 on submit


pascwhale (approved):
This really improves the UX of the application. The changes work perfectly!

PR #462, opened 2022-02-28
Entry create endpoint extension

### LINKED ISSUE #461: Entry create response can take campaign id
## Description
Entry create response should be able to take a campaign id and return some campaign info so that the entry form can be auto completed to an extent.

IanjhPhillips:
## Description
Closes #461
Adds a case to entry and DMentry controller create endpoints that allow them to accept campaign_id as part of the request and pass necessary info in the response to allow for the form to be partially autocompleted.


m-triassi (approved):
We pair programmed this, so im pretty sure its good :)

PR #464, opened 2022-03-01
Character detail to use correct edit entry form

### LINKED ISSUE #428: DM/regular entry are not differentiated in character detail
## Description

Character detail entry table does not differentiate regular and DM entries, we need to open different forms based on the type of the entry.

## Expected Behavior

If the entry is DM, open DM entry form
if the entry is regular, open regular entry form

## Actual Behavior

Only opens regular entry right now

willyyhuang:
Closes #428 

Character detail entry edit to use the correct form in drawer


pascwhale (approved):
It works beautifully!

PR #465, opened 2022-03-01
Add default layout and progress bar

willyyhuang:
This branch adds default layout and progress bar from inertia.

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -30,7 +30,7 @@ i18n.use(initReactI18next) // passes i18n down to react-i18next
 dayjs.extend(localizedFormat)
 dayjs.locale(i18n.language.includes('en') ? 'en' : 'fr-ca')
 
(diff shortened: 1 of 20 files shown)


resources/js/app.js
```diff
@@ -28,8 +30,13 @@ i18n.use(initReactI18next) // passes i18n down to react-i18next
 dayjs.extend(localizedFormat)
 dayjs.locale(i18n.language.includes('en') ? 'en' : 'fr-ca')
 
+InertiaProgress.init()
```
pascwhale, line 33:
```suggestion
InertiaProgress.init({ color: '#8DA57C' })
```

pascwhale: I think using our default green would be better here.

pascwhale (approved):
This is really cool and a significant upgrade to the application's UX. All I ask is that you change the progress bar color to our default green.

PR #468, opened 2022-03-02
Automate Session Entries Job

### LINKED ISSUE #375: Automate event session entries
## Description

When an event session's start time passes, each character as well as the GM of the session must have an entry automatically created for them.

## Possible Implementation
Consider using Laravel jobs

JasonGadoury:
## Description

Closes #375 

This implements the AutomateSessionEntries job that will be executed every 15 minutes. This job scans through sessions to check which have started or passed, and creates an entry for every character & DM participating in them.


JasonGadoury:
Where is the best place to dispatch the job to the queue `AutomateSessionEntries::dispatch();` ? In the documentation and most examples this is done in a controller for which ever model the job is acting upon. It seems like we can simply queue it in the constuct method of the job too...

m-triassi (commented):
Not bad so far, i think you're almost there!

app/Jobs/AutomateSessionEntries.php
```diff
@@ -0,0 +1,82 @@
...
+    public function handle()
+    {
+        //Grab all existing live/past sessions in DB and load characters
+        $sessions = Session::where('start_time', '<=', Carbon::now())
+            ->where('start_time', '>', Carbon::now()->subMinutes(15))
+            ->with('characters')
+            ->get();
+
+        //entry creation input arrays for all characters & dm in $sessions
+        $input_arrays = collect([]);
```
m-triassi, line 48: This variable isnt following naming conventions, it should be `$inputArrays`, camel case, not snake case :)

app/Jobs/AutomateSessionEntries.php
```diff
@@ -0,0 +1,82 @@
...
+                'location' => $session->event->location,
+                'type' => Entry::TYPE_DM,
+            ]));
+        }
+
+        //instantiate all entries with input arrays
+        $event_ids = $input_arrays->pluck('event_id'); //is it possible to use pluck on regular 2d arrays (not collection)
+        $events = Event::whereIn('id', $event_ids)->get();
+        foreach ($events as $event) {
+            $event_entries = $input_arrays->where('event_id', $event->id);
```
m-triassi, line 78: same here, `$eventIds` and `$eventEntries`, in camel case

database/migrations/2022_02_20_075758_create_jobs_table.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::dropIfExists('jobs');
+    }
+}
```
m-triassi, line 36: i believe we only need this migration if the jobs are being tracked via the database, but we're using Redis to en-queue jobs, no?

app/Jobs/AutomateSessionEntries.php
```diff
@@ -0,0 +1,82 @@
...
+                'adventure_id' => $session->adventure_id,
+                'event_id' => $session->event_id,
+                'date_played' => $session->start_time,
+                'location' => $session->event->location,
+                'type' => Entry::TYPE_DM,
+            ]));
+        }
+
+        //instantiate all entries with input arrays
+        $event_ids = $input_arrays->pluck('event_id'); //is it possible to use pluck on regular 2d arrays (not collection)
```
m-triassi, line 75: > is it possible to use pluck on regular 2d arrays (not collection)?

`pluck()` is a function on the `Collection` class, so the variable that you call pluck on should be a collection

app/Jobs/AutomateSessionEntries.php
```diff
@@ -0,0 +1,82 @@
...
+        $sessions = Session::where('start_time', '<=', Carbon::now())
+            ->where('start_time', '>', Carbon::now()->subMinutes(15))
+            ->with('characters')
+            ->get();
+
+        //entry creation input arrays for all characters & dm in $sessions
+        $input_arrays = collect([]);
+        foreach ($sessions as $session) {
+            foreach ($session->characters as $character) {
+                $input_arrays->push(collect([
```
m-triassi, line 51: this call to `collect(...)` adds unnecessary overhead, you can just remove it

app/Jobs/AutomateSessionEntries.php
```diff
@@ -0,0 +1,82 @@
...
+                    'event_id' => $session->event_id,
+                    'dungeon_master_id' => $session->dungeon_master_id,
+                    'date_played' => $session->start_time,
+                    'location' =>$session->event->location,
+                    'type' => Entry::TYPE_GAME,
+                ]));
+            }
+
+            //create DM's entry
+            $input_arrays->push(collect([
```
m-triassi, line 64: same here, no need to the call to `collect(...)`

m-triassi (changes requested):
pretty clever work around there, i'd say the refactor you did is definitely more optimal. Now instead of calling `::create(...)` in a loop, and creating 1 session at a time, you create all the sessions for an event in 1 go. thats substantially less queries to run at a time!

Just a few coding conventions to fix mostly — you can probably take this out of draft btw :)

app/Jobs/AutomateSessionEntries.php
JasonGadoury: True, brain still in python mode.

app/Jobs/AutomateSessionEntries.php
JasonGadoury: I made calls to `collect(...)` because otherwise inputArrays is treated as a regular array and won't let me use collection functions like pluck. It seems to work fine if I only use it to initialize the empty collection. Within the loop you're right it doesn't seem to be necessary. Does that make sense or is there a better way to do that?

database/migrations/2022_02_20_075758_create_jobs_table.php
JasonGadoury: I had done it with database and assumed you would elaborate on replacing it with Redis in your initial review after we spoke about it briefly. My assumption was that since you didn't it was fine as-is but I can make the switch to Redis if it's better.

m-triassi: i see, yeah, I suppose it doesn't matter, this just means that if a developer doesn't have redis running locally they can fall back to the database driver for enqueing jobs, let's leave it

app/Jobs/AutomateSessionEntries.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        //
+    }
```
m-triassi, line 32: you can just remove this empty constructor 
```suggestion
```

app/Jobs/AutomateSessionEntries.php
```diff
@@ -0,0 +1,82 @@
...
+use App\Models\Entry;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+use Carbon\Carbon;
+
+//Remove ShouldBeUnique if we partition the sessions to be handled
```
m-triassi, line 16: is this still relevant? i'm not sure i follow what this means 😅 
```suggestion
```

app/Console/Kernel.php
```diff
@@ -24,7 +25,8 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        // $schedule->command('inspire')->hourly();
+        //Do we have a different preference for how to schedule this job?
```
m-triassi, line 28: nope, this is the correct way to schedule jobs per the docs, you're good!
```suggestion
```

m-triassi (approved):
nice work, solid refactor away from an N+1 query situation, and solid tests. i have some superficial comments here about removing comments and un-needed lines, but im going to approve anyway, you can just commit all the suggestions in a batch, and then merge once the tests run again

PR #469, opened 2022-03-02
User deregistration from event session

### LINKED ISSUE #438: Add destroy function to EventRegistrationController
## Description

Users should be able to leave / un-register from event sessions the registered for. Essentially they should be able to destroy their event registration. Add a function to do that.

image

## Possible Implementation

- Add Destroy method on `EventRegistrationController`, accept just the session ID, and then delete the record that has any of their characters registered for that session. method should use the DELETE verb, ie `Route::delete(...)`

## Out of scope

N/A

N/A

JasonGadoury:
### Description
Closes #438 
Implementation of destroy function in EventRegistrationController to allow user to deregister an owned character from a session. 

### To Test
- Create a session and a character
- Register character for session using `$character->sessions()->attach($session)` ...(could also make a more formal store request passing the event_id, session_id and character_id instead)
- Make call to destroy function passing it the session object as a parameter and the character's id in the request
- Verify if the session was removed from the character's sessions() collection

### DIFF
--- a/app/Http/Controllers/EventRegistrationController.php
+++ b/app/Http/Controllers/EventRegistrationController.php
@@ -7,6 +7,7 @@ use App\Models\Character;
 use App\Models\Event;
 use App\Models\Session;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
 use Illuminate\Validation\UnauthorizedException;
 use Inertia\Inertia;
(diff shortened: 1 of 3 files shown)


IanjhPhillips (approved):
Code looks good and clean, cant see any problems.

Looking into testing this in tinker rn but otherwise should be all good to merge!

PR #474, opened 2022-03-03
Adventures League Import Response Bug Fix

### LINKED ISSUE #424: Importer Bad File Error Screen Bug
## Description

When uploading a bad file to the adventure's league importer, the user is shown an error message that they should not see.

## Expected Behavior
Either the Exception handler or the controller catches the error and redirects the user back

## Actual Behavior
The user is shown this

## Steps to Reproduce

1. Click create a character
2. Select adventures league import
3. Upload a garbage file

IanjhPhillips:
## Description
Closes #424
Previously the adventures league importer and the associated controller were not handling file exceptions correctly.
When the user uploaded a bad file, or a missing file, they were shown a verbose php error. Instead, the controller now redirects them back to the character import page.

## Test
Log in as a user
Navigate to the character import page
Attempt to import from a bad file (any file that is not a valid csv is fine, for example an image of your cat)
Check that the user is redirected back to the same page
Attempt to import a valid file (found in the database/mocks directory of the project)
Check the you are redirected to the created character's page

### DIFF
--- a/app/Exceptions/ImportException.php
+++ b/app/Exceptions/ImportException.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace App\Exceptions;
+
+use Exception;
+
+class ImportException extends Exception
+{
+    public function __construct($message = "Character could not be imported.", $code = 400, Throwable $previous = null)
+    {
+        parent::__construct($message, $code, $previous);
+    }
+}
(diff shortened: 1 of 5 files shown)


m-triassi (approved):
Overall this looks good, so i'll approve, but i remember during a conversation you mentioned wanting to adjust the response bag to not return to the form with the bad file still loaded, so i'll let you decide if you're going to do that or not

m-triassi:
To avoid this getting stale, i'm going to merge this.

IanjhPhillips:
Sounds good! We can make an icebox task to clean the input field on redirected back.

PR #475, opened 2022-03-04
Campaign Kick Endpoint

### LINKED ISSUE #432: Campaign Kick Endpoint
## Description

Campaign owner should be able to kick characters and their corresponding user from their campaign.
The campaign join code should also be  regenerated after a kick request is performed.

## Possible Implementation

Add a destroy function to the campaign registration controller which accepts an array of user ids and removes the user and their characters from the campaign.

RobertoNittolo:
## Description 
Closes #432 
- Add campaign registration destroy route.
- Add campaign registration destroy controller function.
- Add unit tests.

### DIFF
--- a/app/Http/Controllers/CampaignRegistrationController.php
+++ b/app/Http/Controllers/CampaignRegistrationController.php
@@ -54,24 +54,30 @@ class CampaignRegistrationController extends Controller
         return redirect('campaign.index');
     }
 
-    public function destroy(Request $request, Campaign $campaign)
+    public function destroy(Request $request, Campaign $campaign_registration)
     {
         $data = $request->validate([
             'user_id' => 'sometimes|array'
         ]);
 
         $user = Auth::user();
-        $isCampaignOwner = $campaign->users()->where('user_id', $user->id)->first()->pivot->is_owner;
+        $isCampaignOwner = $campaign_registration->users()->where('user_id', $user->id)->first()->pivot->is_owner;
 
-        if ($request->has('user_id') && $isCampaignOwner) {
-            $campaignCharacters = $campaign->characters()->whereIn('user_id', $data['user_id'])->get();
-
-            $campaign->users()->detach($data['user_id']);
-            $campaign->characters()->detach($campaignCharacters);
-
-            $campaign->code = $campaign->generateCode();
-            $campaign->save();
+        if (!$request->has('user_id')) {
+            return redirect()->back()->withErrors(['error' => "You need to specify a user."]);
         }
+
+        if (!$isCampaignOwner) {
+            return redirect()->back()->withErrors(['error' => "You don't have permission to do that."]);
+        }
+
+        $campaignCharacters = $campaign_registration->characters()->whereIn('user_id', $data['user_id'])->get();
+
+        $campaign_registration->users()->detach($data['user_id']);
+        $campaign_registration->characters()->detach($campaignCharacters);
+
+        $campaign_registration->code = $campaign_registration->generateCode();
+        $campaign_registration->save();
 
         return redirect('campaign.index');
     }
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -53,4 +53,26 @@ public function store(Request $request)
...
+
+        if ($request->has('user_id') && $isCampaignOwner) {
+            $campaignCharacters = $campaign->characters()->whereIn('user_id', $data['user_id'])->get();
+
+            $campaign->users()->detach($data['user_id']);
+            $campaign->characters()->detach($campaignCharacters);
+
+            $campaign->code = $campaign->generateCode();
+            $campaign->save();
+        }
```
m-triassi, line 74: pretty good, just missing some error handling here, in cases like this its better to exit early when the state lacks authorization: 
```suggestion
        $user = Auth::user();
        $isCampaignOwner = $campaign->users()->where('user_id', $user->id)->first()->pivot->is_owner;

        if (!$request->has('user_id')) {
            return redirect()->back()->withErrors(['error' => "You need to specify a user."])
        }
        
        if (!$isCampaignOwner) {
            return redirect()->back()->withErrors(['error' => "You don't have permission to do that."])
        }
        
        $campaignCharacters = $campaign->characters()->whereIn('user_id', $data['user_id'])->get();
        
        $campaign->users()->detach($data['user_id']);
        $campaign->characters()->detach($campaignCharacters);

        $campaign->code = $campaign->generateCode();
        $campaign->save();
        
```

tests/Feature/Http/Controllers/CampaignRegistrationControllerTest.php
```diff
@@ -106,4 +106,81 @@ public function store_registers_with_invalid_campaign_code()
...
+        $campaign->characters()->attach($character);
+        $campaign->characters()->attach($character2);
+
+
+        $response = $this->delete(route('campaign-registration.destroy', $campaign), [
+                'user_id' => [$character->user_id, $character2->user_id]
+        ]);
+
+
+
```
m-triassi, line 141: just one new line is fine 😂 
```suggestion

```

m-triassi (changes requested):
pretty good, just a couple changes!

tests/Feature/Http/Controllers/CampaignRegistrationControllerTest.php
RobertoNittolo: lol oops😂

app/Http/Controllers/CampaignRegistrationController.php
RobertoNittolo: I agree, I added the changes!

routes/web.php
```diff
@@ -100,6 +100,9 @@
     Route::resource('campaign-registration', App\Http\Controllers\CampaignRegistrationController::class)
         ->only(['create', 'store']);
 
+    Route::delete('/campaign-registration/{campaign}', [App\Http\Controllers\CampaignRegistrationController::class, 'destroy'])
+        ->name("campaign-registration.destroy");
+
```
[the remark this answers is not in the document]
RobertoNittolo: I agree we don't need that, I'll make the changes you suggested!

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -53,4 +53,32 @@ public function store(Request $request)
 
         return redirect('campaign.index');
     }
+
+    public function destroy(Request $request, Campaign $campaign_registration)
```
m-triassi, line 57: Ah just realized this variable name goes against our code conventions it should be
```suggestion
    public function destroy(Request $request, Campaign $campaignRegistration)
```

Can you rename it everywhere in the method?

m-triassi (changes requested):
one last thing as far as i can tell!

m-triassi (approved):
Looks great!! nice work :)

PR #478, opened 2022-03-05
Connecting Join Form to Backend

### LINKED ISSUE #418: Finish connecting BE to Campaign Join Form
## Description

Join form still imcomplete because it is blocked so moving it to next sprint until index is fully implemented

AnnieTran13:
closes #418 

- Clicking on the join button in the campaign index will open a modal.
- Input code and clicking join will redirect to join form
- After filling out the fields and clicking join, redirected to campaign index with new campaign added

### DIFF
--- a/resources/js/Components/CampaignJoinModal/CampaignJoinModal.tsx
+++ b/resources/js/Components/CampaignJoinModal/CampaignJoinModal.tsx
@@ -33,12 +33,14 @@ type CampaignJoinModalPropType = {
 
 type CampaignJoinModalDataType = {
     code: string | undefined
+    link: string | undefined
 }
 
 const CampaignJoinModal = ({open, onClose, message}: CampaignJoinModalPropType) => {
     const {t} = useTranslation()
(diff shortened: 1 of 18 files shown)


resources/js/Components/CampaignJoinModal/CampaignJoinModal.tsx
```diff
@@ -38,11 +36,10 @@ type CampaignJoinModalDataType = {
 }
 
 const CampaignJoinModal = ({open, onClose, message}: CampaignJoinModalPropType) => {
-    const {data, setData, errors, clearErrors, post} = useForm<CampaignJoinModalDataType>(
-        {
-            code: undefined,
-        },
-    )
+    const {t} = useTranslation()
+    const {data, setData, errors, clearErrors, get} = useForm<CampaignJoinModalDataType>({
```
willyyhuang, line 40: honestly don't think we need `errors` here, it's unlikely we get an error from this, or have you tried entering wrong code?
also i don't think entering an invite link works here either, should it be working?

willyyhuang (approved):

PR #481, opened 2022-03-08
Filter event controller index results

### LINKED ISSUE #437: Filter EventController index results
## Description

Enable the ability to filter the event index by either the event title, description, location, adventure title**s** or by events the user is currently registered for. 

image

## Possible Implementation

- Use the scope Query and filterable functionality added in #436 to reduce the number of results returned. --> accept 1 search parameter for filtererable fields, 1 for the registered toggle.
- Add a call that will load the `league` relationship `->load('league')` on the event.

## Out of scope

N/A

N/A

bcobo341:
## Description
Closes #437 
- Filter event index

### DIFF
--- a/app/Http/Controllers/EventController.php
+++ b/app/Http/Controllers/EventController.php
@@ -16,9 +16,11 @@ class EventController extends Controller
      */
     public function index(Request $request)
     {
-        $events = Event::whereRegistered()
-            ->filtered($request->get('search'))
-            ->get();
+        $events = Event::filtered($request->get('search'));
+        if ($request->has('registered_user') || $request->has('registered_only')) {
+            $events = $events->whereRegistered($request->get('registered_user'));
+        }
+        $events = $events->get();
         $events->load('league');
 
         return Inertia::render('Event/Event', compact('events'));
--- a/app/Http/Controllers/EventController.php
+++ b/app/Http/Controllers/EventController.php
@@ -20,8 +20,7 @@ class EventController extends Controller
         if ($request->has('registered_user') || $request->has('registered_only')) {
             $events = $events->whereRegistered($request->get('registered_user'));
         }
-        $events = $events->get();
-        $events->load('league');
+        $events = $events->with('league')->get();
 
         return Inertia::render('Event/Event', compact('events'));
     }
--- a/app/Http/Controllers/EventController.php
+++ b/app/Http/Controllers/EventController.php
@@ -19,8 +19,8 @@ class EventController extends Controller
         $events = Event::whereRegistered()
             ->filtered($request->get('search'))
             ->get();
-
         $events->load('league');
+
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/EventController.php
```diff
@@ -16,8 +16,11 @@ class EventController extends Controller
      */
     public function index(Request $request)
     {
-        $events = Event::all();
+        $events = Event::whereRegistered()
```
m-triassi, line 19: if you have this here, it means that we're *always* filtering the events by if a user is registered or not. What you'd want to do is check for a flag like `$request->has('registered_only')` or what ever name you want in an if statement, if it has that then you add the `whereRegistered()` call, otherwise don't add it

m-triassi (commented):
So there was an issue with recursive relations loading, i fixed that for you, but while looking at this i noticed that this doesn't capture the exact filtering criteria properly as well

app/Http/Controllers/EventController.php
```diff
@@ -16,8 +16,13 @@ class EventController extends Controller
      */
     public function index(Request $request)
     {
-        $events = Event::all();
+        $events = Event::filtered($request->get('search'));
+        if ($request->has('registered_user') || $request->has('registered_only')) {
+            $events = $events->whereRegistered($request->get('registered_user'));
+        }
+        $events = $events->get();
         $events->load('league');
```
m-triassi, line 24: I recall you saying something about using `->with(...)` instead of `->load()` for this, and you're absolutely right we can do that here instead, and in fact its more efficient: 
```suggestion
        $events = Event::filtered($request->get('search'));
        if ($request->has('registered_user') || $request->has('registered_only')) {
            $events = $events->whereRegistered($request->get('registered_user'));
        }
        $events = $events->with('league')->get();
```

tests/Feature/Http/Controllers/EventControllerTest.php
```diff
@@ -35,15 +37,35 @@ public function setUp(): void
...
+        $sessionFactory = Session::factory()->has($characterFactory);
+        $event = Event::factory()->has($sessionFactory)->create();
+        // Create events that _shouldn't_ be returned
+        Event::factory(3)->has(Session::factory(5))->create();
+        $user = $event->sessions[0]->characters[0]->user;
+
+        $response = $this->actingAs($user)->get(route('event.index', [
+            'search' => $event->title,
+            'registered_only' => true,
+            'registered_user' => $user->id,
```
m-triassi, line 50: what about a test for if `registered_user` is null? i suspect it likely will be null since 99% of the time we're filtering this list for the logged in user, which `whereRegistered()` does automatically

m-triassi (changes requested):
Couple comments for ya!

m-triassi (approved):
Seems about right to me!

PR #483, opened 2022-03-09
Remove react import in favor of new JSX transform

willyyhuang:
This branch is taking advantage of new React version's feature about new JSX transformation. 

Read more: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

Makes the code a little cleaner in general, other than that, using the newest standard is always a good to have.


resources/js/Components/Form/CharacterCreateForm.test.js
```diff
@@ -9,7 +8,7 @@ const factions = [
     'The Harpers',
     'The Order of the Gauntlet',
     'The Emerald Enclave',
-    'The Lords\' Alliance',
+    'The Lords Alliance',
```
pascwhale, line 11: This should not be changed. The faction name is correct. [Source](https://dnd.wizards.com/dungeons-and-dragons/story/factions)

m-triassi (changes requested):
Typo to fix; as an aside, i can appreciate wanting to add nice shiny new react features into the project, lets focus on doing tasks that progress the user stories instead

pascwhale (changes requested):
Everything looks good. I added a comment where you're changing one of the faction names to be incorrect. Otherwise, everything works locally on my machine.

resources/js/Components/Form/CharacterCreateForm.test.js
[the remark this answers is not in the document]
willyyhuang: > that apostrophe wasn't there by accident, the name is supposed to be "The Lords' Alliance"

This does not really matter, it's a mocked prop, the reason why it was removed is because eslint does not appreciate having single quote escaped within single quote, fixed for now but this does not affect anything including the test case.

willyyhuang: See above.

m-triassi (approved):
Ok, this is fine

PR #488, opened 2022-03-10
Fix validation rules for Campaigns

### LINKED ISSUE #485: Campaign character assignment should be optional
## Expected Behavior

When creating a campaign, character should be optional, meaning that the field can be left as `null`

## Actual Behavior

When creating a campaign, it is required.
When editing a campaign, leaving the field `null` will not result an error, rather updates the campaign successfully

m-triassi:
## Description
Closes #485 

Adjusts validation rules to match expected inputs from front end

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -121,7 +121,7 @@ class CampaignController extends Controller
     /**
      * @param \App\Http\Requests\CampaignUpdateRequest $request
      * @param \App\Models\Campaign $campaign
-     * @return \Illuminate\Http\Response
+     * @return \Illuminate\Http\RedirectResponse
      */
     public function update(CampaignUpdateRequest $request, Campaign $campaign)
     {
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -49,12 +49,13 @@ class CampaignController extends Controller
         $search = $data['search'] ?? '';
         $adventures = Adventure::filtered($search)->get(['id', 'title', 'code']);
         $characters = Character::where('user_id', Auth::user()->id)->get();
+
         return Inertia::render('Campaign/Create/CampaignCreate', compact('characters', 'adventures'));
     }
 
     /**
      * @param \App\Http\Requests\CampaignStoreRequest $request
-     * @return \Illuminate\Http\Response
+     * @return \Illuminate\Http\RedirectResponse
      */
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/CampaignController.php
```diff
@@ -143,10 +144,14 @@ public function update(CampaignUpdateRequest $request, Campaign $campaign)
     /**
      * @param \Illuminate\Http\Request $request
      * @param \App\Models\Campaign $campaign
-     * @return \Illuminate\Http\Response
+     * @return \Illuminate\Http\RedirectResponse
      */
     public function destroy(Request $request, Campaign $campaign)
     {
+        if (!$campaign->is_owner) {
```
JasonGadoury, line 151: Could/ should the `getIsOwnerAttribute()` be used here instead of accessing `is_owner` directly? If not, which context is it best suited for

JasonGadoury (approved):
These modifications & additions look good. Have a couple of questions regarding the approach on a couple lines but functionality-wise everything looks fine.

app/Http/Controllers/CampaignController.php
```diff
@@ -102,7 +103,7 @@ public function show(Request $request, Campaign $campaign)
             'campaign' => $campaign,
             'userCharacter' => $userCharacter,
             'characters' => $characters,
-            'adventures' => Adventure::filtered($search)->get(['id', 'title', 'code']),
+            'adventures' => fn () => Adventure::filtered($search)->get(['id', 'title', 'code']),
```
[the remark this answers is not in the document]
m-triassi: from an unassuming observers perspective it would seem like it was working right, but there was actually an issue with _not_ passing a callback here. Intertia has a partial reload function that allows the front end to reload a single specified prop from the backend with new data. For that partial reloading functionality to work though, the prop needs to be passed as a callable so it can evaluate it independently

app/Http/Controllers/CampaignController.php
m-triassi: With [accessors in Laravel](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor), you should always be calling them by their their attribute name instead of calling the function directly. The reason for this is because laravel mostly expects you to access all data about a model through the "magic getter" method from the base eloquent model. Plus its a best practice to access it that way

m-triassi:
Looks like we're good to roll on this one! :) merging! thanks, @JasonGadoury

PR #489, opened 2022-03-11
Fix campaign character duplication bug

### LINKED ISSUE #486: Duplicate campaign on index page when joining the same campaign with different character
## Description

Registering the same campaign with different character will result duplicate entry in campaign index page

## Expected Behavior

Will show up as ONE entry when there are duplicates,
I am also not sure if it is allowed for multiple characters under the same owner to join the same campaign

## Actual Behavior

Duplicate entries in campaign index table

m-triassi:
## Description
Closes #486 

Prevents a user from registering the same character to the same campaign more than once. 
Additionally prevents a user from being registered to a campaign more than once.

### DIFF
--- a/app/Http/Controllers/CampaignRegistrationController.php
+++ b/app/Http/Controllers/CampaignRegistrationController.php
@@ -44,16 +44,23 @@ class CampaignRegistrationController extends Controller
         $campaign = Campaign::where('code', $data['code'])->firstOrFail();
 
         if ($request->has('character_id')) {
-            $user->campaigns()->attach($campaign, ['is_dm' => false, 'is_owner' => false]);
+            $user->campaigns()->syncWithoutDetaching([$campaign->id => ['is_dm' => false, 'is_owner' => false]]);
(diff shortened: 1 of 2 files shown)


JasonGadoury (approved):
Looks great. Tested it locally and the user and character counts stayed the same after a 2nd registration was attempted. I would have expected a failed registration attempt with some error message, however I guess allowing it and using syncWithoutDetatching gives the user a more frictionless experience if that was the intention. 

Edit: My approval was meant to be a comment and I intended to approve after addressing the above. However I still think it's fine as-is.

m-triassi:
I see what you mean, yeah i though about throwing an exception, but i'm starting to rethink if we actually want to be throwing exceptions in controllers now. It's probably preferable to return a response with errors instead. That said, doing that, especially in this context, adds a ton of overhead i'm not sure i want to add — really good points though

m-triassi:
Code climate is being weird and only reporting the diff coverage as 90% even though all the lines are covered, which is making the total coverage show 0.2% drop. Im just going to ignore that

PR #490, opened 2022-03-11
Filter registered sessions on event detail page

### LINKED ISSUE #482: Filter registered sessions on event detail page
## Description

Filter registered sessions on the event details page. Registered sessions under an event are those where a character belonging to the logged in user is included in the list of characters registered to a session.

bcobo341:
## Description
Closes #482 
- Filter registered sessions

### DIFF
--- a/resources/js/Pages/Event/Detail/index.ts
+++ b/resources/js/Pages/Event/Detail/index.ts
@@ -1,1 +1,1 @@
-export {default as CharacterDetail} from './EventDetail'
+export {default as EventDetail} from './EventDetail'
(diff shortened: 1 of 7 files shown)


resources/js/Pages/Event/Detail/index.ts
```diff
@@ -0,0 +1 @@
+export {default as CharacterDetail} from './EventDetail'
```
pascwhale, line 1:
```suggestion
export {default as EventDetail} from './EventDetail'
```

pascwhale (changes requested):
Everything looks good except for on change I added as a suggestion.

pascwhale (approved):
Thank you for addressing the requested change. The frontend files look correct.

m-triassi (approved):
Super clean Backend implementation, nice work!

PR #495, opened 2022-03-13
Fix redirects for campaign registration

### LINKED ISSUE #492: Campaign registration wrong code redirect
## Description

When inputting a wrong code while trying to join a campaign, the page is blank

## Expected Behavior

Should redirect to index page if the code doesn't exist

## Actual Behavior

Goes to the page anyways but typescript errors blocks page from rendering

## Steps to Reproduce

1. Go to campaign index 
2. Click join
3. Enter random code

## Possible Fix

In campaign registration controller, if the code is not valid, redirect to the index page or simply do a reload

# Other

m-triassi:
## Description
Closes #492 

Adjusts redirects when joining or creating campaigns to redirect to the appropriate place and return errors where appropriate

### DIFF
--- a/app/Http/Controllers/CampaignRegistrationController.php
+++ b/app/Http/Controllers/CampaignRegistrationController.php
@@ -12,7 +12,7 @@ class CampaignRegistrationController extends Controller
 {
     /**
      * @param \Illuminate\Http\Request $request
-     * @return \Inertia\Response
+     * @return \Illuminate\Http\RedirectResponse|\Inertia\Response
      */
     public function create(Request $request)
     {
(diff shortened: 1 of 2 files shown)


willyyhuang (approved):
lgtm!

app/Http/Controllers/CampaignRegistrationController.php
```diff
@@ -24,6 +24,10 @@ public function create(Request $request)
 
         $campaign = Campaign::with('characters', 'users')->where('code', $code)->first();
 
+        if (!$campaign) {
+            return redirect()->back()->withErrors(['error' => "The provided campaign code is incorrect."]);
+        }
```
JasonGadoury, line 29: I don't see a test for this, I'm guessing since coverage isn't checked on this PR? Would probably only need 4-5 lines of code.

JasonGadoury (approved):
Looks fine, don't see a test for the added branch. It may or may not be negligible once feat is merged into main, your call. Should be good to merge either way.

app/Http/Controllers/CampaignRegistrationController.php
m-triassi: Just added it :) good  catch!

PR #517, opened 2022-03-14
Hide hard coded content

willyyhuang:


AnnieTran13 (approved):

m-triassi (changes requested):
Instead of commenting things out, can you just delete them instead? if you really need the code later it'll still be available in git

willyyhuang:
> Instead of commenting things out, can you just delete them instead? if you really need the code later it'll still be available in git

yaya okay

m-triassi (approved):
perfect!

PR #527, opened 2022-03-21
GM Entry Creation Bug Fix

### LINKED ISSUE #519: DM entry creation bug
## Description

If a logged in user creates a campaign and assigns a character, when they create a DM entry and select the recently created campaign, then the campaign gets created and appears in the campaign entry table.

## Expected Behavior

The user should not be able to create the DM entry for a campaign where they are not the GM.

## Steps to Reproduce

1. Create campaign and assign character
2. Create DM entry and select the campaign
3. Return to campaign details page

## Possible Fix

Return an error in DMEntryCreate when user selects a campaign where they are not the GM.
OR
Filter the list of campaigns returned in DMEntryCreate in order to only return campaigns where they are GM

# Other

IanjhPhillips:
## Description
Closes #519 
Prevents users from creating GM entries for campaigns in which they are not a GM.

## Changes
Changes the campaigns returned in the GMEntry create endpoint
Also changes the GMEntry auth

### DIFF
--- a/app/Exceptions/GMEntryException.php
+++ b/app/Exceptions/GMEntryException.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace App\Exceptions;
+
+use Exception;
+
+class GMEntryException extends Exception
+{
+    public function __construct($message = "GM Entry Exception", $code = 400)
+    {
+        parent::__construct($message, $code);
+    }
+}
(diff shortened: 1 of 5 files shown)


RobertoNittolo (approved):
Tested the changes locally, I was not able to create a GM entry for a user that is not a GM, as expected! I did notice one thing, that we might be able to change. A non-GM user is still able to access the GM entry page even if they are not a Gm by having the URL, although they are still not able to create the GM entry as expected.
 
`EX: https://localhost/dm-entry/create?campaign_id=4`.

I will approve this PR as it does exactly what is expected of it, and we can discuss if we want to block the user from accessing the page altogether. Great Job!

IanjhPhillips:
@RobertoNittolo 

Yea I made sure to account for that which is why the controller does a check as well. Even though they can access the page they wont be able to do the request.

PR #528, opened 2022-03-22
Use prepareForValidation instead of inline controller logic

### LINKED ISSUE #503: Refactor Request validators to use prepareForValidation function
## Description

Previously we changed the way controllers handled expected input because of the search field. It would be preferable to not have to use array access for these and have them work as expected

## Possible Implementation

Use the `prepareForValidation()` function that can be overridden in the request validator and prepare the incoming data to match the `ENTITYNAME_id` pattern instead.

## Out of scope

N/A

N/A

m-triassi:
## Description
Closes #503 
Refactors many of the request validators to use a generic base class which handled merging the appropriate model ID into the request before validating it. 

This removes the need for repeated inline logic within controllers.

## How to test
There should be effectively no perceivable difference in the way entry as well as campaign storage / updating work what-so-ever. Thus simply test the original features.


JasonGadoury (approved):
Looks much cleaner than before. All features work fine as well. Great refactor!

PR #529, opened 2022-03-24
Create Cypress Tests for League Admin Event Crud

### LINKED ISSUE #439: Cypress Tests for League Admin Event Control
## Description

Write E2E Cypress tests for #363 League Admin Event Control.

## Possible Implementation

- View League Events
- Create a League Event
- Edit a League Event
- Delete a League Event

IanjhPhillips:
## Description
Closes #439
Creates league admin user in db seeders
Uses new user in new league admin crud tests


m-triassi (approved):
The only real problem i see here is the ~35 seconds of idle time at the start of the video — but i suppose that's unavoidable since the seeder has to run. Otherwise seems pretty good to me!

pascwhale (approved):
Everything looks good to me. This is an optional change but we could maybe use the same `seederLogin` function and use an optional parameter to determine which account we log into.

willyyhuang (approved):

PR #530, opened 2022-03-25
Fix rating cascade delete error on campaign delete

### LINKED ISSUE #512: Deleting Campaign with Ratings Throws Exception
## Description

Deleting a campaign that has dm ratings throws an exception

## Expected Behavior

Campaign should be deleted successfully 

## Actual Behavior

Query exception: integrity constraint violation is thrown

## Steps to Reproduce

1. Log in with test account
2. Create campaign, (choose "None" as character)
3. Logout of test account
4. Login with different user
5. Join the campaign created in step 2 with different user
6. Create an Entry within that campaign, including the submission of a dm rating
7. Log back in with user who created campaign
8. Delete campaign

## Context
Particular edge case for campaign deletion failing

## Possible Fix
Fix foreign key constraints between ratings, entries, and campaigns tables

# Other

m-triassi:
## Description
Closes #512 

Adds soft deletes to campaigns to allow for easily deleting / archiving of campaigns without violating cascade constraints.

This solution avoids messy n+1 detach logic that would have to be added to the destroy logic in the controller. 

## How to test
Simply attempt to delete a campaign which has entries / ratings attached to it.

### DIFF
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -5,6 +5,7 @@ namespace App\Models;
 use App\Models\Traits\Filterable;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Facades\Auth;
 
 class Campaign extends Model
@@ -12,6 +13,7 @@ class Campaign extends Model
     use \Backpack\CRUD\app\Models\Traits\CrudTrait;
     use HasFactory;
     use Filterable;
+    use SoftDeletes;
 
     /**
      * The attributes that are mass assignable.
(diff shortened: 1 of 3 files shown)


JasonGadoury (approved):
Tested it, works great. I wonder if over time we will have a lot of deleted campaigns and therefore unnecessary extra baggage in our table since we're using soft delete. Might affect performance down the line unless deleting campaigns is rare. In any case, this works fine! Good stuff

m-triassi:
@JasonGadoury Yeah i was considering that too, but i figure it'll be easier to do add a job later that cleans up very old campaigns or something. 

The reason i went with soft deletes was that otherwise we'd have a loop over all the entries from a campaign to detach them, which is slow. i didn't want our deletes speed to scale inversely with campaign entries. 

Otherwise i might consider a hacky solution in the future that directly delete the entries on the pivot table or something

PR #531, opened 2022-03-29
Update Reward accessor to return Null

### LINKED ISSUE #446: Reward choice set a '-' in list view when it's null
## Description

The reward choice is set to `-` which is not an option for front-end, therefore when sending request to update, backend will think that reward choice is not null

## Expected Behavior

Return null in the list, front-end will format if there is no reward choice

## Actual Behavior

Return `-`, when editing, front-end set edit data for reward as `-`, therefore updating fails

## Steps to Reproduce

1. 
2. 
3. 

## Additional Notes

Please fix this cypress test when fixing this.

m-triassi:
## Description
Closes #446 

Minor formatting change, update the computed reward field to return null and instead depend on FE to format / display a specific text.

### DIFF
--- a/app/Models/Entry.php
+++ b/app/Models/Entry.php
@@ -118,12 +118,8 @@ class Entry extends Model
 
     public function getRewardAttribute()
     {
-        if ($this->type != self::TYPE_DM) {
+        if ($this->type != self::TYPE_DM || is_null($this->character_id)) {
             return null;
-        }
-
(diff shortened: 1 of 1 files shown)


bcobo341 (approved):
This is good!

IanjhPhillips (approved):
Super simple change that does what its supposed to. Empty reward choices still show up as you would expect on my local environment so I cant see any problem with this.

Good work!

PR #532, opened 2022-03-29
Fix DM entry display bug on campaign details

### LINKED ISSUE #501: DM entry not getting deleted properly
## Description

After creating DM entry through campaign details page, then deleting that DM entry in the DM entry index page. The entry will not be deleted in the entries table in the campaign details page.

## Expected Behavior

The DM entry should be deleted.

## Actual Behavior

The DM entry is deleted in the DM entry index but not in the campaigns entry table.

## Steps to Reproduce

1. Create a DM entry through campaign's detail page
2. Go to DM entry index (through characters tab)
3. Delete the DM entry just created
4. Return to the campaign's detail page and the entry is still there

## Possible Fix

# Other

m-triassi:
## Description
Closes #501 

Adjusts the query returning entries for a given campaign to instead load entries via the campaign model rather than the character model. Update the front end to reflect this change as well. Add missing types for entry on campaign as well.

## How to Test
1. Create or join a campaign as the DM
2. Create a new entry
3. Observe that the entry now shows in the entry table on the campaign details page
4. Observe that you can delete the entry which causes it to disappear. 
5. Observe that the entry table still shows up for other player characters in the campaign.

### DIFF
--- a/app/Http/Controllers/CampaignController.php
+++ b/app/Http/Controllers/CampaignController.php
@@ -9,6 +9,8 @@ use App\Models\Entry;
 use App\Models\User;
 use App\Models\Character;
 use App\Models\Adventure;
+use Illuminate\Database\Eloquent\Relations\HasMany;
+use Illuminate\Database\Query\Builder;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
 use Inertia\Inertia;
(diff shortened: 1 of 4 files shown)


willyyhuang (approved):
LGTM!

IanjhPhillips (approved):
Everything looks good here and I see some refactors mixed in as well.

Behaves as expected in the bug report so I think it should be good to merge!

PR #533, opened 2022-03-29
Remove unused controllers / tests / views in code base

### LINKED ISSUE #502: Backend Route Cleanup
## Description

There are some routes that have been auto generated at the beginning of the project that have since not been used. These routes and controller functions should be removed so that malicious users cannot send requests to the endpoints.

m-triassi:
## Description
Close #502 
A large amount of (insecure) code that was added during the Laravel blueprint initialization of the project still remained in the code base. This removes it.

Deletes all non-essential Controller functions, associated tests, and the views that were never used. 

The only exception is the trade controller, which instead has been disabled during all instances except if in testing.


bcobo341 (commented):
I don't see any errors, however, I would like someone else to look into this PR before it is approved. But looks correct to me!

IanjhPhillips (approved):
Cant think of anything missing. Tested the different features as much as i could think to do as I saw the things you deleted so I dont think anything is broken with this.

Also, the cypress tests seem to think that all the features still work, so thats good.

PR #534, opened 2022-03-30
Performance Testing

RobertoNittolo:
## Description 
- Locust Controller changes
  - add the cleanUp method to be used once the performance test has concluded. It will delete all remaining characters and the API token of the locust user. 
  - Modify the delete character method to only delete the first character
- Locust Script changes(not on this PR/Repo)
- Run Locust Script on UAT and analyze the performance(worked well)
- Run performance tests on the production server


RobertoNittolo:
We are done with this PR, we did not need to merge it in since these changes would have needed to be removed afterward as they are changes that are specifically related to performance testing. We were able to test these changes by making a staging environment, with the changes.

PR #540, opened 2022-04-02
Cypress test for Campaign Entry Creation

### LINKED ISSUE #445: Cypress Tests for Campaign Entry Creation
## Description

Write E2E Cypress tests for #252 Campaign Entry Creation.

## Possible Implementation

- Check creation of campaigns for other members

JasonGadoury:
### Description
Closes #445 

Cypress test for Campaign Entry Creation
Creates entry from character enrolled in campaign and checks from another user's perspective if entry appears in CampaignDetail.

This test was added to the original campaign_enrollment.js test suite to avoid duplicating the joining portion at the start.

### DIFF
--- a/cypress/integration/Campaign Management/campaign_enrollment_and_entry_creation.js
+++ b/cypress/integration/Campaign Management/campaign_enrollment_and_entry_creation.js
@@ -1,8 +1,10 @@
-describe('Campaign Enrollment Test Suite', () => {
+describe('Campaign Enrollment and Entry Creation Test Suite', () => {
     let campaign_title
     let code
     let campaign_exploration_link
     let campaign_exploration_id
+    const levelup = 2
+
     before(() => {
         cy.refreshDatabase()
         cy.seed('FastSeeder')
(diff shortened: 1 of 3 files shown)


pascwhale (approved):
Hey @JasonGadoury, good work! Thank you for using the `data-cy` attribute for specific components. I agree that adding your test in the campaign enrollment suite was smart. This does avoid code duplication,

cypress/integration/Campaign Management/campaign_enrollment_and_entry_creation.js
```diff
@@ -51,6 +53,31 @@ describe('Campaign Enrollment Test Suite', () => {
...
+
+        cy.contains('button', 'Entry')
+            .click()
+            .then(() => {
+                cy.get('#levels').clear().type(levelup)
+                cy.contains('button', 'Continue').click()
+                cy.contains('button', 'Create').click()
+                cy.url().should('include', campaign_exploration_id)
+            })
+            .then(() => {
```
willyyhuang, line 71: is there a reason why we are chaining `.then()` here? all of these are sequential events that do not depend on each other

JasonGadoury: The `.then()` segment  will only succeed if the entry is created in the lines right before. Otherwise campaign detail will have no entries.

willyyhuang: ok

willyyhuang (approved):

PR #541, opened 2022-04-02
perf: #539 call cy.login instead of using UI to log into default test…

### LINKED ISSUE #539: Replace Cypress Login Calls with Laravel Login
## Description

Speed up the process of logging into the default test user's account during Cypress tests.

## Possible Implementation

Replace login action for Cypress tests with a simple `cy.login({id: 1})` call to log into the default test user without having to use the UI.

pascwhale:
Closes #539

## Description
Use `cy.login({id: 1})` call to log into the default user instead of using the UI.


pascwhale:
These changes will not be used. See #539.

PR #542, opened 2022-04-03
Hotfix: add session seats to backpack

m-triassi:
## Description
Adds the missing `seats` column to laravel backpack. This allows Site and league admins to set the available amount of characters that can register for a session. Also adds the missing `language` field

Also adds some more realistic values to factories, so that characters generated now have real races and classes

Hotfix as this is a last minute bug that came up.

Adds These fields:  
image

### DIFF
--- a/app/Http/Controllers/Admin/SessionCrudController.php
+++ b/app/Http/Controllers/Admin/SessionCrudController.php
@@ -61,6 +61,7 @@ class SessionCrudController extends CrudController
         CRUD::column('adventure_id');
         CRUD::column('dungeonMaster');
         CRUD::column('table');
+        CRUD::column('seats');
         CRUD::column('start_time');
         CRUD::column('end_time');
     }
@@ -109,6 +110,8 @@ class SessionCrudController extends CrudController
         CRUD::field('adventure_id');
         CRUD::field('dungeonMaster');
         CRUD::field('table');
+        CRUD::field('seats')->default(8);
+        CRUD::field('language')->type('enum');
         CRUD::field('start_time');
         CRUD::field('end_time');
     }
--- a/database/factories/CharacterFactory.php
+++ b/database/factories/CharacterFactory.php
@@ -23,11 +23,38 @@ class CharacterFactory extends Factory
      */
     public function definition()
     {
+        $races = [
+            "Dwarf",
+            "Elf",
+            "Halfling",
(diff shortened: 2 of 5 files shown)


app/Http/Controllers/Admin/SessionCrudController.php
```diff
@@ -109,6 +110,8 @@ protected function setupCreateOperation()
         CRUD::field('adventure_id');
         CRUD::field('dungeonMaster');
         CRUD::field('table');
+        CRUD::field('seats')->default(8);
```
IanjhPhillips, line 113: big table

database/factories/CharacterFactory.php
```diff
@@ -23,11 +23,38 @@ class CharacterFactory extends Factory
...
+            "Halfling",
+            "Human",
+            "Dragonborn",
+            "Gnome",
+            "Half-Elf",
+            "Half-Orc",
+            "Tiefling",
+        ];
+
+        $classes = [
```
IanjhPhillips, line 38: Artificer is AL too now

IanjhPhillips (changes requested):
Looks good overall! I would just add the Artificer class. Otherwise i think its good to go

app/Http/Controllers/Admin/SessionCrudController.php
IanjhPhillips: (not a change request)

database/factories/CharacterFactory.php
IanjhPhillips: You can ignore this actually lol I blanked on the fact that this is for the seeder so doesnt ultimately matter.

IanjhPhillips (approved):

app/Http/Controllers/Admin/SessionCrudController.php
m-triassi: yeah it the typical maximum size that DnD MTL use at their events, most tables dont go that big but i figured it was the number that made the most sense

database/factories/CharacterFactory.php
m-triassi: True, but its not in the SRD, and technically not allowed to be used publically on non WoTC systems — im sure it wouldn't matter though since this factory is really only used for testing and local environments, so ill go ahead and add that :)

PR #545, opened 2022-04-14
Hide language selector when user is logged in

### LINKED ISSUE #544: Hide language selector on the menu if the user is logged in
## Description

Language selector is shown after logging in

willyyhuang:
Closes #544 

Hide language selector when user is logged in, the property language is bind to user's account


m-triassi (approved):
Ah perfect, yeah that toggle was a bit confusing

PR #549, opened 2022-06-15
Add webkit scrollbar to the app

willyyhuang:
This branch simply adds webkit scrollbar to the supported browsers, make overflow scroll looks nicer, this selector targets all the MUI components.


willyyhuang:
it's mostly for autocomplete component!

